PR #2, opened 2020-04-24
PT-139 | Tunnistamo authentication

jorilindell:
1. Add redux store to the project to store application state
2. Add react-toastify to show error/success messages
3. Implement log in/out using Tunnistamo
4. Add instructions to set up Tunnistamo locally


.env
```diff
@@ -1 +1,5 @@
 SASS_PATH=./src/styles
+
+REACT_APP_OIDC_AUTHORITY="http://tunnistamo-backend:8000"
```
badrange, line 3: I recommend a solution that doesn't require a local tunnistamo by default. In kukkuu we have an .env.development like this: 

```REACT_APP_API_URI="https://kukkuu.test.kuva.hel.ninja/graphql"
REACT_APP_OIDC_AUTHORITY="https://tunnistamo.test.kuva.hel.ninja"```

And if we want to run it against a local tunnistamo we create a `.env.development.local` with the local overrides

File: src/domain/auth/authenticate.ts [hunk 2]
```diff
@@ -0,0 +1,77 @@
...
+  }
+};
+
+export const getApiToken = (accessToken: string): StoreThunk => async (
+  dispatch: Function
+) => {
+  try {
+    dispatch(startFetchingToken());
+
+    const res: AxiosResponse<ApiTokenResponse> = await axios.post(
```
Nizar-Rahme, line 55: For me `axios.get` served the purpose here.

src/domain/auth/oidcCallback/OidcCallback.tsx
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  return (
+    <Container>
+      <CallbackComponent
+        successCallback={onSuccess}
+        errorCallback={onError}
+        userManager={userManager}
+      >
+        <p>{t('authentication.redirect.text')}</p>
```
Nizar-Rahme, line 42: The translation here is missing, IMO, a loading spinner looks nicer, or better yet the main page layout.

PR #6, opened 2020-05-08
Image selector component implementation

Sartonon:
Problem: seems like errors persist even when image is set to empty string:
```
  image: Yup.string(),
  photographer: Yup.string().when('image', {
    is: (image) => image,
    then: Yup.string().required(VALIDATION_MESSAGE_KEYS.STRING_REQUIRED),
    otherwise: Yup.string(),
  }),
  imageAltText: Yup.string().when('image', {
    is: (image) => image,
    then: Yup.string().required(VALIDATION_MESSAGE_KEYS.STRING_REQUIRED),
    otherwise: Yup.string(),
  }),
```


src/domain/event/eventForm/EventForm.tsx
```diff
@@ -170,4 +184,96 @@ const EventForm = () => {
...
+      <img
+        className={styles.eventImage}
+        src={image}
+        alt={t('eventForm.basicInfo.eventImageAltText')}
+      />
+    </div>
+  );
+};
+
+const DeleteImageButton: React.FC<{
```
jorilindell, line 265: This could more generic component so we could use it in other forms as well. Maybe it could accept label and aria-label as props so it would be easier to re-use. 

Could be renamed e.g. DeleteButton

src/domain/event/eventForm/EventForm.tsx
```diff
@@ -170,4 +184,96 @@ const EventForm = () => {
...
+  onClick: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+}> = ({ onClick }) => {
+  const { t } = useTranslation();
+  return (
+    <button className={styles.deleteImageButton} onClick={onClick}>
```
jorilindell, line 270: aria-label needed to have better accessibility support

src/domain/event/eventForm/ValidationSchema.ts
```diff
@@ -29,4 +29,15 @@ export default Yup.object().shape({
   providerContactInfo: Yup.object().shape({
     email: Yup.string().email(VALIDATION_MESSAGE_KEYS.EMAIL),
   }),
+  image: Yup.string(),
+  photographer: Yup.string().when('image', {
```
jorilindell, line 33: Maybe validateField method could be used to validate photographer and imageAltText fields after deleting image
https://jaredpalmer.com/formik/docs/guides/validation

src/domain/event/eventForm/EventForm.tsx
```diff
@@ -170,4 +184,96 @@ const EventForm = () => {
   );
 };
 
+const ImageSelectedForm: React.FC<{
+  image: string;
```
jorilindell, line 188: Could be moved to own file. ImageSelectedForm is little bit misleading name because this component is not form

src/domain/event/eventForm/EventForm.tsx
```diff
@@ -170,4 +184,96 @@ const EventForm = () => {
...
+            />
+          </FormGroup>
+        </div>
+      </div>
+    </>
+  );
+};
+
+const SelectImageForm: React.FC = () => {
+  const { t } = useTranslation();
```
jorilindell, line 232: Could be moved to own file. SelectImageForm is little bit misleading name because this component is not form

src/domain/event/eventForm/EventForm.tsx
```diff
@@ -170,4 +184,96 @@ const EventForm = () => {
...
+      </div>
+      <div className={styles.imageInstructionsText}>
+        {t('eventForm.basicInfo.descriptionAddImage')}
+      </div>
+    </div>
+  );
+};
+
+const ImagePreview: React.FC<{ image: string }> = ({ image }) => {
+  const { t } = useTranslation();
```
jorilindell, line 252: Could be moved to own file

src/domain/event/eventForm/EventForm.tsx
Sartonon: Good point. Any suggestions for a better name? ImageSelectedFormPart?

jorilindell: That's ok to me. Let's try to use same convention in the future as well to keep the codebase consistent

Sartonon: 👍

PR #20, opened 2020-05-18
PT-166 | Save event and page to show event details

jorilindell:
- Save event feature
- Event details page

Only fields which can be saved at the moment are included to new event page and event details page


src/domain/event/CreateEventPage.tsx
```diff
@@ -1,19 +1,89 @@
...
+                },
+              };
+
+              try {
+                const data = await createEvent({
+                  variables: { event: payload },
+                });
+                const id =
+                  data.data?.addEventMutation?.response?.body?.id || '666';
+                history.push(
```
Sartonon, line 79: I thiknk you can leave /${locale} off here:

 history.push(ROUTES.EVENT_DETAILS.replace(':id', id));

jorilindell: This is good one. Let's use it also in the future.

PR #21, opened 2020-05-19
PT-167, PT-168 | Edit event page

jorilindell:
1. Page to edit existing event
2. Support different language versions of the event
3. Show warning if there are unsaved changes on the form

### DIFF
--- a/src/domain/event/CreateEventPage.tsx
+++ b/src/domain/event/CreateEventPage.tsx
@@ -7,7 +7,6 @@ import useLocale from '../../hooks/useLocale';
 import Container from '../app/layout/Container';
 import PageWrapper from '../app/layout/PageWrapper';
 import { ROUTES } from '../app/routes/constants';
-import EditEventButtons from './editEventButtons/EditEventButtons';
 import EventForm from './eventForm/EventForm';
 import styles from './eventPage.module.scss';
 import { getEventPayload } from './utils';
@@ -24,11 +23,6 @@ const CreateEventPage: React.FC = () => {
     <PageWrapper title="createEvent.pageTitle">
       <Container>
         <div className={styles.eventPage}>
-          <EditEventButtons
-            onClickLanguage={setSelectedLanguage}
-            selectedLanguage={selectedLanguage}
-          />
-          <h1>{t('createEvent.title')}</h1>
           <EventForm
             onSubmit={async (values) => {
               try {
@@ -44,6 +38,9 @@ const CreateEventPage: React.FC = () => {
                 // Check apolloClient to see error handling
               }
             }}
+            selectedLanguage={selectedLanguage}
+            setSelectedLanguage={setSelectedLanguage}
+            title={t('createEvent.title')}
           />
         </div>
       </Container>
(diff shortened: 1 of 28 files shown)


src/domain/event/CreateEventPage.tsx
```diff
@@ -2,85 +2,45 @@ import React from 'react';
...
-  // TODO: Get event language which user has selected when implemented
-  const language = locale;
 
   return (
     <PageWrapper title="createEvent.pageTitle">
       <Container>
         <div className={styles.eventPage}>
-          <h1>{t('createEvent.title')}</h1>
           <EventForm
             onSubmit={async (values) => {
```
Sartonon, line 27: maybe move this logic to its own function from the jsx

src/domain/event/EditEventPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  return (
+    <PageWrapper title="editEvent.pageTitle">
+      <LoadingSpinner isLoading={loading}>
+        <Container>
+          <div className={styles.eventPage}>
+            <EventForm
+              eventData={eventData}
+              initialValues={initialValues}
+              onSubmit={async (values) => {
```
Sartonon, line 72: maybe move this to its own function

src/domain/event/eventDetailsButtons/EventDetailsButtons.tsx
```diff
@@ -0,0 +1,63 @@
...
+const EventDetailsButtons: React.FC<Props> = ({
+  eventData,
+  onClickLanguage,
+  selectedLanguage,
+}) => {
+  const { t } = useTranslation();
+  const history = useHistory();
+  const eventId = eventData.event?.id || '';
+
+  const moveToEventList = () => {
```
Sartonon, line 29: would goToEventList be better name?

src/domain/event/eventDetailsButtons/EventDetailsButtons.tsx
```diff
@@ -0,0 +1,63 @@
...
+    history.push(ROUTES.HOME);
+  };
+
+  const moveToEditPage = () => {
```
Sartonon, line 33: would goToEditPage be better name?

src/domain/event/utils.ts
```diff
@@ -13,3 +18,70 @@ export const getEventPlaceholderImage = (id: string): string => {
...
   return EVENT_PLACEHOLDER_IMAGES[index];
 };
+
+/**
+ * Get first availanle language of an event
+ * @param {object} eventData
+ * @return {string}
+ */
+export const getFirstAvailableLanguage = (eventData: EventQuery): Language => {
+  if (eventData.event?.name.en) return SUPPORT_LANGUAGES.EN;
```
Sartonon, line 28: Is this the correct order? should FI be first?

src/domain/event/utils.ts
```diff
@@ -13,3 +18,70 @@ export const getEventPlaceholderImage = (id: string): string => {
 
   return EVENT_PLACEHOLDER_IMAGES[index];
 };
+
+/**
+ * Get first availanle language of an event
```
Sartonon, line 23: btw why do we need these docs?

src/domain/event/utils.ts
Sartonon: maybe should be same order as in the language buttons?

Screenshot 2020-05-19 at 11 34 39

src/domain/event/utils.ts
jorilindell: Maybe we don't. I'm just used to write them

PR #24, opened 2020-05-20
Delete redundant bottom margin from actions buttons because padding w…

Sartonon:
…as added to the container

### DIFF
--- a/src/domain/occurrence/eventOccurrenceForm/eventOccurrenceForm.module.scss
+++ b/src/domain/occurrence/eventOccurrenceForm/eventOccurrenceForm.module.scss
@@ -52,7 +52,7 @@
 
   .formActions {
     display: flex;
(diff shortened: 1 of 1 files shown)


src/domain/occurrence/eventOccurrenceForm/eventOccurrenceForm.module.scss
```diff
@@ -52,7 +52,7 @@
 
   .formActions {
     display: flex;
-    margin: $largeMargin 0;
+    margin-top: $largeMargin 0;
```
jorilindell, line 55: I guess this should be 
`margin-top: $largeMargin;`

Sartonon: yes :) Will fix later

PR #25, opened 2020-05-20
PT-212 | Upload images 

jorilindell:
1. Support uploading image from event form
2. Implement component to display image information
3. Fine-tune remove button styles


src/domain/event/EditEventPage.tsx
```diff
@@ -53,14 +59,57 @@ const EditEventPage: React.FC = () => {
...
+              id: eventData?.event?.id || '',
+              ...getEventPayload(values, selectedLanguage),
+            },
           },
-        },
-      });
+        })
+      );
+
+      const shouldSaveImage =
```
Sartonon, line 75: this could be moved to its own functions and be called in the if statetement to make this submit function little bit cleaner

src/domain/event/eventForm/ImageSelectedFormPart.tsx
```diff
@@ -20,42 +20,52 @@ const ImageSelectedFormPart: React.FC<{
...
-      <ImagePreview image={image} />
+      <ImagePreview id={imageId} />
       <div className={styles.imageSelectRow}>
         <div className={styles.imageSelect}>
           <DeleteButton
             onClick={() => {
-              ['image', 'imageAltText', 'photographer'].forEach((field) => {
-                setFieldValue(field, '');
-              });
+              ['image', 'imageAltText', 'imagePhotographerName'].forEach(
```
Sartonon, line 48: could be moved to own function e.g. clearImageFormFields or something :)

PR #27, opened 2020-05-25
Timepicker accepts value as Date and update occurrence form according …

Sartonon:
…to latest design

- Datepicker now accepts value as Date

- Update eventOccurenceForm to latest design

### DIFF
--- a/src/common/components/datepicker/Datepicker.tsx
+++ b/src/common/components/datepicker/Datepicker.tsx
@@ -28,8 +28,8 @@ type Props = {
   invalidText?: string;
   labelText?: string;
   onBlur: () => void;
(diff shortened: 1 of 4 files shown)


src/common/components/datepicker/Datepicker.tsx
```diff
@@ -21,15 +21,17 @@ import DatepickerContext from './datepickerContext';
 import Month from './Month';
 import MonthNavButton from './MonthNavButton';
 
+const dateRegex = /^\d\d\.\d\d\.\d\d\d\d$/;
```
jorilindell, line 24: Maybe this could be written
`/^\d{2}\.\d{2}\.\d{4}$/`

PR #29, opened 2020-05-25
PT-204 | Support line breaks on event description

jorilindell:


src/common/components/textWithLineBreaks/__tests__/TextWithLineBreaks.test.tsx
```diff
@@ -0,0 +1,11 @@
+import { shallow } from 'enzyme';
```
Sartonon, line 1: why enzyme?

src/common/components/textWithLineBreaks/__tests__/TextWithLineBreaks.test.tsx
```diff
@@ -0,0 +1,11 @@
...
+import * as React from 'react';
```
Sartonon, line 2: why import * as React from 'react'?

jorilindell: eslint is showing following error "React must be in scope when using JSX" if React is not imported

src/common/components/textWithLineBreaks/__tests__/TextWithLineBreaks.test.tsx
jorilindell: Replaced with @testing-library/react

PR #31, opened 2020-05-27
PT-215 | Create/edit occurrence page

jorilindell:
1. Open existing occurrence by clicking occurrence table row 
2. Implement page to create new occurrence
3. Implement page to edit existing occurrence
4. Correctly reset value on DatePicker and TimePicker components
5. Scroll to the back top when pathname is changed

### DIFF
--- a/src/domain/occurrence/EditOccurrencePage.tsx
+++ b/src/domain/occurrence/EditOccurrencePage.tsx
@@ -1,7 +1,150 @@
+import { Button } from 'hds-react';
 import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useHistory, useParams } from 'react-router';
 
-const EditOccurrencePage = () => {
-  return <div>TODO: EDIT OCCURRENCE PAGE</div>;
+import LoadingSpinner from '../../common/components/loadingSpinner/LoadingSpinner';
+import {
+  useEditOccurrenceMutation,
+  useEventQuery,
+  useOccurrenceQuery,
+} from '../../generated/graphql';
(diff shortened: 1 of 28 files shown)


src/domain/occurrence/EditOccurrencePage.tsx
```diff
@@ -1,7 +1,155 @@
...
+  };
+
+  const goToOccurrencesPage = () => {
+    history.push(`/${locale}${ROUTES.OCCURRENCES.replace(':id', eventId)}`);
+  };
+
+  const getPayload = (values: OccurrenceFormFields) => {
+    return {
+      id: occurrenceId,
+      ...getOccurrencePayload(
```
Sartonon, line 62: Consider passing arguments as an object and use destructuring in the function to avoid having to put so many arguments in the correct order

src/domain/occurrence/EditOccurrencePage.tsx
```diff
@@ -1,7 +1,155 @@
...
+                </h1>
+                <Button variant="secondary" onClick={goToEventDetailsPage}>
+                  {t('editOccurrence.buttonShowEventInfo')}
+                </Button>
+              </div>
+              <EventOccurrenceForm
+                eventId={eventId}
+                formTitle={t('editOccurrence.formTitle')}
+                initialValues={{
+                  date: occurrenceData.occurrence?.startTime
```
Sartonon, line 120: Maybe add getInitialValues function and use that

Sartonon: to make the jsx little bit cleaner

PR #35, opened 2020-05-29
Add tests fro datepicker

Sartonon:


src/hooks/useDropdownKeyboardNavigation.tsx
```diff
@@ -68,7 +68,7 @@ const useDropdownKeyboardNavigation = ({
           break;
         case 'ArrowDown':
           if (isInitialNavigation && typeof initialFocusedIndex === 'number') {
-            focusOption('down', initialFocusedIndex);
+            focusOption('down', initialFocusedIndex - 1);
```
jorilindell, line 71: What is going to happen if initialFocusedIndex is 0?

src/common/components/datepicker/TimesList.tsx
```diff
@@ -111,7 +111,7 @@ const TimeItem: React.FC<TimeItemProps> = ({
   // scroll focused element into view
   React.useEffect(() => {
     if (focused) {
-      buttonRef.current?.scrollIntoView({
+      buttonRef.current?.scrollIntoView?.({
```
jorilindell, line 114: I'm only curious here, but what is the reason for ?-mark in scrollIntoView?

Sartonon: jsdom seems to not support that one and errors out. Maybe need add comment about that?

src/hooks/useDropdownKeyboardNavigation.tsx
Sartonon: First down key will focus the first item. At least seems to work correctly for times list and language selector

jorilindell: Ok, great

src/common/components/datepicker/TimesList.tsx
jorilindell: That's something that I was expecting. Maybe comment could be helpful. It was not obvious to me when I saw this at first time.

PR #36, opened 2020-06-01
Add error/not found pages

Sartonon:
## Description

Add ErrorPage component and use it in error situations

### Closes
**[PT-214](https://helsinkisolutionoffice.atlassian.net/browse/PT-214):** 
**[PT-213](https://helsinkisolutionoffice.atlassian.net/browse/PT-213):** 

## Screenshots

Screenshot 2020-06-01 at 9 13 30
Screenshot 2020-06-01 at 9 13 12

### DIFF
--- a/src/domain/errorPage/ErrorPage.module.scss
+++ b/src/domain/errorPage/ErrorPage.module.scss
@@ -1,0 +1,33 @@
+@import 'variables';
+
+.container {
+  --header-fontsize: var(--fontsize-h-2);
+}
+
+.container {
+  text-align: center;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding-top: var(--spacing-4-xl);
+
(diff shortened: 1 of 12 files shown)


src/domain/errorPage/ErrorPage.module.scss
```diff
@@ -0,0 +1,33 @@
...
+    width: 3rem;
+  }
+
+  h1 {
+    font-size: var(--header-fontsize);
+    margin: var(--spacing-l) 0;
+  }
+
+  .description {
+    max-width: 550px;
```
jorilindell, line 24: Does this break layout on mobile devices? I'm not sure

Sartonon: I don't thinks so, it is only max-width. But the container needs some left and right padding for the mobile width

PR #38, opened 2020-06-01
Prevent editing past events add error toasts

Sartonon:
## Description

- Show error page when trying to edit past event

- Hide edit event button when event is in the past

- Show error messages on edit event and create event pages

### Closes
**[PT-232](https://helsinkisolutionoffice.atlassian.net/browse/PT-232):**

### DIFF
--- a/src/domain/app/i18n/fi.json
+++ b/src/domain/app/i18n/fi.json
@@ -64,7 +64,8 @@
   },
   "createEvent": {
     "pageTitle": "Uusi tapahtuma",
-    "title": "Uusi tapahtuma"
+    "title": "Uusi tapahtuma",
+    "error": "Tapahtuma tallentaminen epäonnistui"
   },
   "editEvent": {
     "buttons": {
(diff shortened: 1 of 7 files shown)


src/domain/event/eventDetailsButtons/EventDetailsButtons.tsx
```diff
@@ -28,6 +29,11 @@ const EventDetailsButtons: React.FC<Props> = ({
   const history = useHistory();
   const eventId = eventData.event?.id || '';
 
+  const isFutureEvent = () =>
```
jorilindell, line 32: Maybe you could re-use function from event/utils.js

src/domain/app/i18n/fi.json
```diff
@@ -64,15 +64,19 @@
...
   },
   "editEvent": {
     "buttons": {
       "buttonBack": "Tapahtumat",
       "textDirty": "Sivulla on tallentamattomia muutoksia"
     },
     "pageTitle": "Muokkaa tapahtumaa",
-    "title": "Muokkaa tapahtumaa"
+    "title": "Muokkaa tapahtumaa",
+    "error": "Tapahtuma tallentaminen epäonnistui",
```
jorilindell, line 77: Maybe we could change this to "Tapahtuman tallentaminen epäonnistui"

src/domain/app/i18n/fi.json
```diff
@@ -64,15 +64,19 @@
   },
   "createEvent": {
     "pageTitle": "Uusi tapahtuma",
-    "title": "Uusi tapahtuma"
+    "title": "Uusi tapahtuma",
+    "error": "Tapahtuma tallentaminen epäonnistui"
```
jorilindell, line 68: Maybe we could change this to "Tapahtuman tallentaminen epäonnistui"

PR #42, opened 2020-06-05
PT-197 | Event list page

jorilindell:
Description
Implement event list page

Issues
Closes
PT-197:


src/domain/event/utils.ts
```diff
@@ -131,3 +134,20 @@ export const isFutureEvent = (eventData: EventQuery | undefined) =>
...
+  let hasComingItems = false;
+
+  forEach(event.pEvent?.occurrences.edges, (edge) => {
+    if (edge?.node?.startTime && isFuture(new Date(edge?.node?.startTime))) {
+      hasComingItems = true;
+      return false;
+    }
+  });
+
+  return Boolean(hasComingItems);
```
Sartonon, line 152: Maybe a little bit redundant to have Boolean() here because hasComingItems is already a boolean and the name implies that also

src/domain/event/utils.ts
```diff
@@ -131,3 +134,20 @@ export const isFutureEvent = (eventData: EventQuery | undefined) =>
 
 export const isEditableEvent = (eventData: EventQuery | undefined) =>
   isFutureEvent(eventData);
+
+export const hasOccurrences = (event: EventFieldsFragment) => {
+  return Boolean(event.pEvent?.occurrences.edges.length);
+};
+
+export const hasComingOccurrences = (event: EventFieldsFragment) => {
```
Sartonon, line 142: I think adding boolean return type for this function would be good

src/domain/event/utils.ts
```diff
@@ -131,3 +134,20 @@ export const isFutureEvent = (eventData: EventQuery | undefined) =>
 
 export const isEditableEvent = (eventData: EventQuery | undefined) =>
   isFutureEvent(eventData);
+
+export const hasOccurrences = (event: EventFieldsFragment) => {
```
Sartonon, line 138: I think adding boolean return type for this function would be good

src/domain/event/utils.ts
Sartonon: If you change/refactor this it will tell you if you are not returning boolean anymore

src/domain/events/EventsPage.tsx
```diff
@@ -1,19 +1,140 @@
...
 const EventsPage = () => {
+  const [inputValue, setInputValue] = React.useState('');
+  const searchValue = useDebounce(inputValue, 100);
   const { t } = useTranslation();
+  const locale = useLocale();
   const history = useHistory();
+  const [isLoadingMore, setIsLoadingMore] = React.useState(false);
+
+  const { data, loading, fetchMore } = useEventsQuery({
+    errorPolicy: 'ignore',
```
Sartonon, line 75: Does API return error here in some situations? Maybe backend shouldn't return error in some situations that it does now

src/utils/getPageFromUrl.ts
```diff
@@ -0,0 +1,5 @@
+export default function getPageFromUrl(url: string) {
```
Sartonon, line 1: here you could also explicitly tell the return type

Sartonon: also maybe getPageNumberFromUrl would be little bit more descriptive name for this function. I was little bit confused what this function actually returns before reading the implementation

src/domain/events/EventsPage.tsx
jorilindell: Yes, we are now creating events locally and to http://palvelutarjotin-api.test.kuva.hel.ninja/. If pEvent doesn't exist in used environment Apollo will throw an error without errorPolicy: "ignore"

PR #60, opened 2020-07-09
PT-323 | Enrolment table

jorilindell:
### Description
Enrolment table without actions dropdown

### Closes
PT-323

### Screenshot
Screenshot 2020-07-09 at 19 44 51

### Additional info
Can be tested on http://localhost:3000/en/events/palvelutarjotin:afzii4ixmu/occurrences/T2NjdXJyZW5jZU5vZGU6Nw==


src/domain/occurrence/enrolmentTable/additionalInfo/AdditionalInfo.tsx
```diff
@@ -0,0 +1,52 @@
...
+    enrolment.person?.language || enrolment.studyGroup.person.language;
+  const studyGroupPersonInfo = [
+    enrolment.studyGroup.person.phoneNumber,
+    enrolment.studyGroup.person.emailAddress,
+  ].filter((e) => e);
+  const personInfo = [
+    enrolment.person?.phoneNumber,
+    enrolment.person?.emailAddress,
+  ].filter((e) => e);
+  const extraNeeds = enrolment.studyGroup.extraNeeds;
```
Sartonon, line 25: Would something like getEnrolmentField be a good fit here?

src/domain/occurrence/enrolmentTable/EnrolmentTable.tsx
```diff
@@ -0,0 +1,178 @@
...
+      accessor: (row: EnrolmentFieldsFragment) => row,
+      Cell: ({
+        row,
+      }: {
+        value: EnrolmentFieldsFragment;
+        row: {
+          getToggleRowExpandedProps: () => void;
+          isExpanded: boolean;
+        };
+      }) => {
```
Sartonon, line 128: I know that this was taken from my code :) but what do you think, should we make a new type to make this more readable?

jorilindell: Tried to make more generic type for the column cell

src/domain/occurrence/enrolmentTable/additionalInfo/AdditionalInfo.tsx
jorilindell: Added helper function

PR #74, opened 2020-08-11
PT-354: CreateEventPage tests

Sartonon:
## Description

Add tests fro CreateEventPage

### DIFF
--- a/src/domain/event/CreateEventPage.tsx
+++ b/src/domain/event/CreateEventPage.tsx
@@ -117,7 +117,9 @@ const CreateEventPage: React.FC = () => {
       });
     } catch (e) {
       // TODO: Improve error handling when API returns more informative errors
-      console.log(e);
+      if (process.env.NODE_ENV === 'test') {
+        console.log(e);
+      }
       toast(t('createEvent.error'), {
         type: toast.TYPE.ERROR,
       });
--- a/src/domain/event/CreateEventPage.tsx
+++ b/src/domain/event/CreateEventPage.tsx
@@ -117,6 +117,7 @@ const CreateEventPage: React.FC = () => {
       });
     } catch (e) {
       // TODO: Improve error handling when API returns more informative errors
+      console.log(e);
       toast(t('createEvent.error'), {
         type: toast.TYPE.ERROR,
       });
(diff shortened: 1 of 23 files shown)


src/domain/event/__tests__/CreateEventPage.test.tsx
```diff
@@ -0,0 +1,406 @@
...
+});
+
+const mocks = [
+  {
+    request: {
+      query: MyProfileDocument,
+    },
+    result: {
+      ...profileResponse,
+    },
```
jorilindell, line 106: Maybe this could be:
`result: profileResponse,`

src/domain/event/__tests__/CreateEventPage.test.tsx
```diff
@@ -0,0 +1,406 @@
...
+
+  const modal = within(screen.getByRole('dialog', { hidden: true }));
+
+  const cancelButton = modal.getByRole('button', {
+    name: 'Peruuta',
+    hidden: true,
+  });
+  userEvent.click(cancelButton);
+
+  expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
```
jorilindell, line 230: Maybe testing confirmation modal could be moved to own test?

src/domain/event/CreateEventPage.tsx
```diff
@@ -117,6 +117,7 @@ const CreateEventPage: React.FC = () => {
       });
     } catch (e) {
       // TODO: Improve error handling when API returns more informative errors
+      console.log(e);
```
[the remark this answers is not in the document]
Sartonon: maybe should add check if it is test environment. Hard to debug tests if there is error and it is not logged

src/domain/event/__tests__/CreateEventPage.test.tsx
Sartonon: yeah

src/domain/event/__tests__/CreateEventPage.test.tsx
Sartonon: Do you mean in this file? I agree, maybe that would be better

Sartonon:
Jest complains about these .json files that there are duplicates. Maybe __mocks__ folder is not the best place for these response files

PR #82, opened 2020-08-18
PT-373 Test/occurrence details

Sartonon:
## Description

- Add render helper method for tests

- Add tests for occurrence details

### Closes
**[PT-373](https://helsinkisolutionoffice.atlassian.net/browse/PT-373):**


src/domain/occurrence/__tests__/OccurrenceDetailsPage.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { ROUTES } from '../../app/routes/constants';
+import OccurrenceDetailsPage from '../OccurrenceDetailsPage';
+import {
+  eventResult,
+  occurrenceResult,
+  placeResult,
+  venueResult,
+} from './mocks';
+
+// expect.extend(toHaveNoViolations);
```
jorilindell, line 25: You can import 'jest-axe/extend-expect' on src/setupTests.ts so you don't have to extend expect object for each test file. e.g. https://github.com/City-of-Helsinki/events-helsinki-ui/blob/develop/src/setupTests.ts

src/domain/occurrence/__tests__/OccurrenceDetailsPage.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+      query: VenueDocument,
+      variables: {
+        id: 'tprek:15376',
+      },
+    },
+    result: venueResult,
+  },
+];
+
+// uncomment this when enrolment table is accessible (labels for checkboxes)
```
jorilindell, line 74: This is really minor. Maybe you could add "TODO:" at the beginning of the comment so it's easier to find in the future

src/domain/occurrence/__tests__/mocks.ts
```diff
@@ -0,0 +1,769 @@
+export const eventResult = {
```
jorilindell, line 1: Maybe you could explicitly add types to these mock responses? It might help if the data structure of the graphql query responses changes at some point

src/utils/testUtils.tsx
```diff
@@ -0,0 +1,86 @@
...
+            {children}
+          </Route>
+        </Router>
+      </MockedProvider>
+    </Provider>
+  );
+
+  const renderResult = render(ui, { wrapper: Wrapper });
+  return { ...renderResult, history };
+};
```
jorilindell, line 65: These are really helpful. I have to add similar ones to Tapahtumat Helsinki project

src/domain/occurrence/__tests__/OccurrenceDetailsPage.test.tsx
Sartonon: yeah I will make that change :)

src/domain/occurrence/__tests__/OccurrenceDetailsPage.test.tsx
Sartonon: Yes you are right :)

src/domain/occurrence/__tests__/mocks.ts
Sartonon: Good point

PR #87, opened 2020-09-06
Fix tests and update context query

Sartonon:
- Refactor modals to user new linkedEvent field

- fix tests


jorilindell (approved):
Looks good

PR #88, opened 2020-09-09
PT-437 | Add event end time when updating event

Sartonon:
## Description

- Added endTime to the mutation input when updating event

### Closes
**[PT-437](https://helsinkisolutionoffice.atlassian.net/browse/PT-437):**


jorilindell (approved):
Good

PR #92, opened 2020-09-15
PT-460 | Fix custom message bug when approving and declining

Sartonon:
## Description

- Fix custom message bug when approving and declining

### Closes
**[PT-460](https://helsinkisolutionoffice.atlassian.net/browse/PT-460):**


jorilindell (approved):
Looks good

PR #97, opened 2020-09-22
PT-475 | Separate queries for different event categories

Sartonon:
## Description

- Have separate load more buttons in events page and different queries for each category

### Closes
**[PT-475](https://helsinkisolutionoffice.atlassian.net/browse/PT-475):** 

## Screenshots

Screenshot 2020-09-22 at 12 42 36

### DIFF
--- a/src/domain/events/EventsPage.tsx
+++ b/src/domain/events/EventsPage.tsx
@@ -46,11 +46,11 @@ const EventsPage: React.FC = () => {
   };
 
   const {
-    data: eventsData,
-    loading: loadingEvents,
-    isLoadingMore: isLoadingMoreEvents,
-    hasNextPage: eventsHasNextPage,
-    fetchMore: fetchMoreEvents,
+    data: upcomingEventsData,
+    loading: loadingUpcomingEvents,
+    isLoadingMore: isLoadingMoreUpcomingEvents,
+    hasNextPage: upcomingEventsHasNextPage,
+    fetchMore: fetchMoreUpcomingEvents,
   } = useEventsQueryHelper({
     errorPolicy: 'ignore',
     variables: {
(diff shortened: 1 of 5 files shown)


src/domain/events/EventsPage.tsx
```diff
@@ -139,50 +141,67 @@ const EventsPage: React.FC = () => {
...
                   goToEventOccurrencesPage={goToEventOccurrencesPage}
                 />
+                {pastEventsHasNextPage && (
+                  <ShowMoreButton
+                    loading={loadingMorePastEvents}
+                    onClick={fetchMorePastEvents}
+                  />
+                )}
               </>
             )}
```
jorilindell, line 204: Maybe you could have separate component to show each event list to avoid duplicate code

Sartonon: Thought about that but I don't know if there is a huge benefit because the duplication happens only twice here (events with upcoming occurrences has little bit different structure). And adding new abstraction level (new component) might make it little bit harder to understand.

Don't know... maybe it would be little bit easier to read. I will add new component :)

jorilindell: I undestand. I almost didn't even comment about it. 

It would be nice to move event fetch hooks inside that component. But I know the loading state is the reason to keep them on this level

Sartonon: Hmm also moving fetching logic inside a event category component crossed my mind but I can't remember why I thought it would not be good idea.

Now that I look the code again I feel like there is no reason not to make a new component that has the fetching logic inside it, just pass query variables as prop? 🤔

jorilindell: Yeah, passing query variables and title should be enough

Sartonon: I was just thinking that maybe that component could be used for the upcoming occurrences (just don't pass title prop). But that is when the loading state becomes issue.

Maybe I have to ask the designer how we want to handle the loading states in the first place :)

jorilindell (approved):
Very good.

PR #107, opened 2020-10-01
Improve test performance

Sartonon:
Improve test performance by configuring defaultHidden to true.

[More info](https://testing-library.com/docs/dom-testing-library/api-queries#byrole)

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "build": "react-scripts build",
     "lint": "eslint --ext js,ts,tsx src",
     "test": "react-scripts test --env=jest-environment-jsdom-sixteen",
-    "test:changed": "npm run test -- -o --watchAll=false",
+    "test:changed": "npm test -- -o --watchAll=false",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -69,7 +69,7 @@
     "build": "react-scripts build",
     "lint": "eslint --ext js,ts,tsx src",
     "test": "react-scripts test --env=jest-environment-jsdom-sixteen",
-    "test:changed": "npm run test -- -o --watchAll=false",
+    "test:changed": "npm test -- -o --watchAll=false",
```
jorilindell, line 72: Maybe `yarn test -- -o --watchAll=false` to be consistent

jorilindell (approved):
This is nice one. The difference seems to be significant. e.g Datepicker.test.tsx 20s vs. 80s

Sartonon:
> This is nice one. The difference seems to be significant. e.g Datepicker.test.tsx 20s vs. 80s

Yeah. getByRole does some expensive checking that is not that important.

package.json
Sartonon: updated

PR #110, opened 2020-10-08
Accessibility improvements

Sartonon:
## Description

- Accessibility improvements

### Related
**[PT-510](https://helsinkisolutionoffice.atlassian.net/browse/PT-510):**

### DIFF
--- a/src/domain/occurrence/enrolmentTable/EnrolmentTable.tsx
+++ b/src/domain/occurrence/enrolmentTable/EnrolmentTable.tsx
@@ -93,6 +93,9 @@ const EnrolmentTable: React.FC<Props> = ({
           id={`${id}_select-all_checkbox`}
           checked={isAllSelected}
           onChange={isAllSelected ? unselectAll : selectAll}
+          aria-label={t(
+            'occurrenceDetails.enrolmentTable.labelSelectAllEnrolments'
+          )}
         />
       ),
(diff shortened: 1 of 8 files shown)


src/domain/occurrence/enrolmentTable/EnrolmentTable.tsx
```diff
@@ -162,6 +173,7 @@ const EnrolmentTable: React.FC<Props> = ({
               'occurrenceDetails.enrolmentTable.showEnrolmentDetails'
             )}
             {...row.getToggleRowExpandedProps()}
+            aria-expanded={row.isExpanded ? true : false}
```
jorilindell, line 176: isExpanded already sounds like a boolean value.

jorilindell (approved):
Looks good

src/domain/occurrence/enrolmentTable/EnrolmentTable.tsx
Sartonon: yeah, for some reason isExpanded is undefined when closed, so had to do this

jorilindell: Ok, makes sense

Sartonon: But maybe comment would be nice here

PR #111, opened 2020-10-12
PT-408 | Fix event date validation when editing

Sartonon:
## Description

- Use different validation for date when editing event

### Closes
**[PT-408](https://helsinkisolutionoffice.atlassian.net/browse/PT-408):**


jorilindell (approved):
Looks good

PR #113, opened 2020-10-16
PT-530 | Prevent editing event that has been published

Sartonon:
## Description

- Prevent editing events that have been published

### Closes
**[PT-530](https://helsinkisolutionoffice.atlassian.net/browse/PT-530):**

### DIFF
--- a/src/domain/event/EditEventPage.tsx
+++ b/src/domain/event/EditEventPage.tsx
@@ -28,6 +28,7 @@ import {
   getEventPayload,
   getEventVenueDescription,
   getFirstAvailableLanguage,
+  isEditableEvent,
 } from './utils';
 
 const EditEventPage: React.FC = () => {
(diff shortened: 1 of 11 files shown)


src/domain/event/EditEventPage.tsx
```diff
@@ -201,7 +202,7 @@ const EditEventPage: React.FC = () => {
       <LoadingSpinner isLoading={loading}>
         {!!eventData ? (
           <>
-            {isEditableEvent(eventData) ? (
+            {isEditableEvent ? (
```
jorilindell, line 205: I would still use isEditableEvent(eventData) function, but change the internal logic. Just to keep maintaining little bit easier. Now ```const isEditableEvent =
    eventData?.event?.publicationStatus === PUBLICATION_STATUS.DRAFT; ``` is duplicated

src/domain/event/eventDetailsButtons/EventDetailsButtons.tsx
```diff
@@ -57,7 +59,7 @@ const EventDetailsButtons: React.FC<Props> = ({
         onClick={onClickLanguage}
         selectedLanguage={selectedLanguage}
       />
-      {isEditableEvent(eventData) && (
+      {isEditable && (
```
jorilindell, line 62: I would use isEditableEvent helper function but change it's internal logic

Sartonon: Good catch. Dunno what I was doing :D

jorilindell (approved):
Very good!

PR #117, opened 2020-10-23
PT-526 | Event wizard step 2

Sartonon:
## Description

- Add second step to event wizard (occurrences)

### Closes
**[PT-526](https://helsinkisolutionoffice.atlassian.net/browse/PT-526):**


src/domain/occurrence/CreateOccurrencePage.tsx
```diff
@@ -179,4 +229,33 @@ const CreateOccurrencePage: React.FC = () => {
...
+            endsAt: initialEndsAt,
+          }
+        : defaultInitialValues;
+    }
+    return defaultInitialValues;
+  }, [initialDate, initialEndsAt, initialStartsAt]);
+
+  return initialFormValues;
+};
+
```
Sartonon, line 260: wasn't quite sure what would be the cleanest way to handle these initial values... ended up with this :D

src/domain/event/CreateEventPage.tsx
```diff
@@ -119,7 +122,14 @@ const CreateEventPage: React.FC = () => {
...
       history.push({
         pathname: `/${locale}${ROUTES.CREATE_OCCURRENCE.replace(':id', id)}`,
+        search: firstOccurrencePrefilledValuesToQuery(values),
       });
+
+      // TODO: think if there is better way to handle this scrollTop.
+      // at the moment <ScrollToTop /> component ignores
+      // '/events/:id/occurrences/create' because it matches one of the ignored paths:
+      // '/events/:id/occurrences/:occurrenceId'
+      window.scrollTo(0, 0);
```
jorilindell, line 132: Could you add new prop to ScrollToTop where you define array of routes that should alway scroll to top. In that case you could /events/:id/occurrences/:occurrenceId to ignored routes and /events/:id/occurrences/create to the routes (I don't know what could be suitable name for that prop) that should always scroll to top

src/domain/occurrence/CreateOccurrencePage.tsx
```diff
@@ -179,4 +229,33 @@ const CreateOccurrencePage: React.FC = () => {
   );
 };
 
+const useInitialFormValues = () => {
```
jorilindell, line 232: Is this same hook also needed on edit page? If yes, you could move this to separate file

src/domain/occurrence/EditOccurrencePage.tsx
```diff
@@ -152,6 +167,19 @@ const EditOccurrencePage: React.FC = () => {
...
+  const handleDeleteOccurrence = async (
+    occurrence: OccurrenceFieldsFragment
+  ) => {
+    try {
+      await deleteOccurrence({ variables: { input: { id: occurrence.id } } });
+      refetchEvent();
+    } catch (e) {
+      toast(t('occurrences.deleteError'), {
+        type: toast.TYPE.ERROR,
+      });
```
jorilindell, line 179: You could also write this line as
```
toast.error(t('occurrences.deleteError'));
```
In that case could can test it easier with Jest (If you want to test the error case). e.g. `toast.error = jest.fn()`

src/hooks/useQuery.ts
```diff
@@ -1,5 +1,12 @@
...
 export const useQuery = () => {
-  return new URLSearchParams(useLocation().search);
+  const search = useLocation().search;
+
+  const query = React.useMemo(() => {
+    return new URLSearchParams(search);
+  }, [search]);
+
+  return query;
 };
```
jorilindell, line 12: Maybe you could rename this hook e.g. useSearchParams. At first time when I read this PR through I wasn't sure what it does

jorilindell (approved):
Looks good. I added couple of comments but they are not deal breakers to me

src/domain/occurrence/CreateOccurrencePage.tsx
Sartonon: did some changes to this already for the next step

src/domain/occurrence/CreateOccurrencePage.tsx
Sartonon: at least not yet, I will move it if it is needed. But for now only create occurrence page needs initialValues from query params

jorilindell: Ok, in that case there is no need to do anything for this

src/hooks/useQuery.ts
Sartonon: renamed

src/domain/occurrence/EditOccurrencePage.tsx
Sartonon: fixed

src/domain/event/CreateEventPage.tsx
Sartonon: added new prop to ScrollToTop component and use that to handle the scrolling

src/domain/occurrence/CreateOccurrencePage.tsx
```diff
@@ -135,16 +158,31 @@ const CreateOccurrencePage: React.FC = () => {
...
+  const handleDeleteOccurrence = async (
+    occurrence: OccurrenceFieldsFragment
+  ) => {
+    try {
+      await deleteOccurrence({ variables: { input: { id: occurrence.id } } });
+      refetchEvent();
+    } catch (e) {
+      toast(t('occurrences.deleteError'), {
+        type: toast.TYPE.ERROR,
+      });
```
jorilindell, line 170: Sorry, I missed this at first time. Maybe could use `toast.error(t('occurrences.deleteError'))` also here

src/domain/app/ScrollToTop.tsx
```diff
@@ -4,27 +4,39 @@ import { matchPath, useLocation } from 'react-router';
...
+        })
+      ),
+    [locale, pathname]
+  );
 
-    if (!isIgnoredPath) {
-      window.scrollTo(0, 0);
+  React.useEffect(() => {
+    if (isMatch(forceScrollToTopPaths)) {
+      debugger;
```
jorilindell, line 28: Is this line still needed?

jorilindell (approved):
Looks good. Added some really minor comments

src/domain/app/ScrollToTop.tsx
Sartonon: nope :D

PR #122, opened 2020-11-11
PT-582 | Initialize different catgories on edit event page

Sartonon:
## Description

- Initialize different catgories on edit event page

### Closes
**[PT-582](https://helsinkisolutionoffice.atlassian.net/browse/PT-582):**


package.json
```diff
@@ -122,10 +122,10 @@
...
   "devDependencies": {
     "@apollo/react-testing": "^3.1.4",
-    "@graphql-codegen/cli": "^1.13.3",
-    "@graphql-codegen/typescript": "^1.13.3",
-    "@graphql-codegen/typescript-operations": "^1.13.3",
-    "@graphql-codegen/typescript-react-apollo": "^1.13.3",
+    "@graphql-codegen/cli": "1.13.3",
+    "@graphql-codegen/typescript": "1.13.3",
+    "@graphql-codegen/typescript-operations": "1.13.3",
+    "@graphql-codegen/typescript-react-apollo": "1.13.3",
```
Sartonon, line 128: If i add "^" back, it installs wrong version and codegen won't work with current script

jorilindell: Ok, Better to use strict version in that case. Maybe all codegen related libraries could be updated at some point.

jorilindell (approved):
Looks good

package.json
Sartonon: yeah

PR #125, opened 2020-12-01
PT-632 fix: Event starting day cannot be the current day

nenonja:
## Description
Allow current day to be valid event starting day. Also changed error message text.
### Closes
**[PT-632](https://helsinkisolutionoffice.atlassian.net/browse/PT-632):** 
### Automated tests
Few component tests added for testing yesterday date and current date
## Screenshots
Note! gif recorded on first of december, so it is the current day in it.
date

### DIFF
--- a/src/common/components/textInput/InputWrapper.tsx
+++ b/src/common/components/textInput/InputWrapper.tsx
@@ -77,7 +77,7 @@ const InputWrapper: FC<InputWrapperProps> = React.forwardRef<
       )}
       <div className={classNames(styles.inputWrapper)}>{children}</div>
       {helperText && (
-        <div className={styles.helperText} id={`${id}-helper`}>
+        <div className={styles.helperText} id={`${id}-helper`} role="alert">
           {helperText}
         </div>
       )}
--- a/src/common/components/textInput/InputWrapper.tsx
+++ b/src/common/components/textInput/InputWrapper.tsx
@@ -76,7 +76,11 @@ const InputWrapper: FC<InputWrapperProps> = React.forwardRef<
         </Tooltip>
       )}
       <div className={classNames(styles.inputWrapper)}>{children}</div>
-      {helperText && <div className={styles.helperText}>{helperText}</div>}
+      {helperText && (
+        <div className={styles.helperText} role="alert">
+          {helperText}
+        </div>
+      )}
     </div>
   )
 );
--- a/src/common/components/textInput/InputWrapper.tsx
+++ b/src/common/components/textInput/InputWrapper.tsx
@@ -77,7 +77,7 @@ const InputWrapper: FC<InputWrapperProps> = React.forwardRef<
       )}
       <div className={classNames(styles.inputWrapper)}>{children}</div>
       {helperText && (
-        <div className={styles.helperText} role="alert">
+        <div className={styles.helperText} id={`${id}-helper`}>
           {helperText}
         </div>
       )}
--- a/src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
+++ b/src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
@@ -1,5 +1,4 @@
 import { MockedResponse } from '@apollo/react-testing';
-import { fireEvent } from '@testing-library/react';
 import { addDays, format, formatISO } from 'date-fns';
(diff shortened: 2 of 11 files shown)


Sartonon:
Looking very good. Great job!

src/domain/occurrence/eventOccurrenceForm/ValidationSchema.ts
```diff
@@ -1,8 +1,7 @@
-import isBefore from 'date-fns/isBefore';
-import isFuture from 'date-fns/isFuture';
-import parseDate from 'date-fns/parse';
+import { isBefore, isToday, parse as parseDate } from 'date-fns';
```
Sartonon, line 1: Is this going to increase the bundle size? These functions were imported line by line for a reason but I'm not sure id date-fns is tree shakeable

src/common/components/textInput/InputWrapper.tsx
```diff
@@ -76,7 +76,11 @@ const InputWrapper: FC<InputWrapperProps> = React.forwardRef<
         </Tooltip>
       )}
       <div className={classNames(styles.inputWrapper)}>{children}</div>
-      {helperText && <div className={styles.helperText}>{helperText}</div>}
+      {helperText && (
+        <div className={styles.helperText} role="alert">
```
nenonja, line 80: `role="alert"` added for accessibilty + easier testing

src/domain/occurrence/eventOccurrenceForm/ValidationSchema.ts
Sartonon: + isToday seems to be unused

src/common/components/textInput/InputWrapper.tsx
Sartonon: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

"The alert role is used to communicate an important and usually time-sensitive message to the user."

The problem here is that this helperText element could appear to many input at the same time, I'm not sure if that is  good user experience with screen readers. Also helperText element is used for other than errors so we would have elements with role="alert" that are not really alert messages

src/domain/occurrence/eventOccurrenceForm/ValidationSchema.ts
Sartonon: maybe we should make sure that we have imported these individually everywhere and test the bundle sizde with:

https://create-react-app.dev/docs/analyzing-the-bundle-size/

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -395,6 +397,9 @@ test('today is valid event start day', async () => {
   fireEvent.blur(dateInput);
   await waitFor(() => {
     expect(dateInput).toBeValid();
+    expect(dateInput).not.toHaveAttribute('aria-describedby');
+    expect(
+      screen.queryByText(messages.form.validation.date.mustNotInThePast)
+    ).not.toBeInTheDocument();
```
Sartonon: This is pretty minor but I don't recommend putting many assertions into waitFor.  You could only wait for dateInput to be valid and then leave rest of the assertions out of the waitFor (below). If 1 of those assertions would fail, the you have to always wait for the timeout to give you the error but there is no reason to wait for those after the first assertion.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -368,10 +369,11 @@ test('yesterday is not valid event start day', async () => {
   fireEvent.blur(dateInput);
   await waitFor(() => {
     expect(dateInput).toBeInvalid();
+    expect(dateInput).toHaveAttribute('aria-describedby');
+    expect(
+      screen.queryByText(messages.form.validation.date.mustNotInThePast)
+    ).toBeInTheDocument();
```
Sartonon: This is pretty minor but I don't recommend putting many assertions into waitFor. You could only wait for dateInput to be valid and then leave rest of the assertions out of the waitFor (below). If 1 of those assertions would fail, the you have to always wait for the timeout to give you the error but there is no reason to wait for those after the first assertion.

Sartonon (approved):
Great :)

PR #130, opened 2020-12-30
PT-642/PT-643: Add tests to GitHub actions

nenonja:
## Issues :bug:
- Add lint and unit tests to ci-workflow
- fix lint tests
- set correct environment vars for unit tests
- browser-tests workflow removed and moved into review-workflow: tests are run as acceptance tests against review einvironment
- browser-tests added as smoke tests for staged and production. environment (can be tested after merge)
### Closes :no_good_woman:
https://helsinkisolutionoffice.atlassian.net/browse/PT-642
https://helsinkisolutionoffice.atlassian.net/browse/PT-643


.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
+      - uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: Install dependencies
+        run: yarn --prefer-offline --frozen-lockfile --check-files
```
Sartonon, line 59: I would like to know why are these flags needed.

.github/workflows/review.yml
```diff
@@ -34,13 +34,47 @@ jobs:
...
+      - name: Setup kubectl
+        run: |
+          echo "${{ env.KUBECONFIG_RAW }}" > $(pwd)/kubeconfig
+          echo "KUBECONFIG=$(pwd)/kubeconfig" >> $GITHUB_ENV
+        shell: bash
+      - name: Get Review Deploy URL
+        id: deploy-url
+        run: |
+          DEPLOY_URL=$(kubectl get ingress -n "${{ env.K8S_NAMESPACE }}" -o jsonpath='{.items[0].spec.rules[0].host}')
+          echo "BROWSER_TESTS_LOCAL_ENV_URL=https://$DEPLOY_URL" >> $GITHUB_ENV
```
Sartonon, line 55: So this sets the env variable for browser tests? What does `>> $GITHUB_ENV` actually do?

.github/workflows/staging.yml
```diff
@@ -49,3 +49,35 @@ jobs:
...
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: Install dependencies
+        run: yarn --prefer-offline --frozen-lockfile --check-files --production=false
+      - name: Run TestCafe tests
+        run: yarn browser-test:ci
+        env:
+          BROWSER_TESTS_LOCAL_ENV_URL: https://${{ secrets.ENVIRONMENT_URL_STAGING }}
```
Sartonon, line 77: how is this used?

src/icons/icons.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import { storiesOf } from '@storybook/react';
-import path from 'path';
-import React from 'react';
+import * as path from 'path';
```
Sartonon, line 2: is this correct?

.github/workflows/staging.yml
```diff
@@ -49,3 +49,35 @@ jobs:
           K8S_NAMESPACE: ${{ secrets.K8S_NAMESPACE_STAGING }}
           ENVIRONMENT_URL: https://${{ secrets.ENVIRONMENT_URL_STAGING }}
           DOCKER_IMAGE_NAME: ${{ env.REPO_NAME }}-staging
+  acceptance-tests:
+    runs-on: windows-latest
+    needs: staging
+    name: Acceprance tests for staging
```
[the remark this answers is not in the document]
nenonja: fixed

.github/workflows/ci.yml
nenonja: https://classic.yarnpkg.com/en/docs/cli/install/
Technically --frozen-lockfile will check that the yarn.lock file in version control matches to the same created by install:
`If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag.`

`yarn install --check-files
Verifies that already installed files in node_modules did not get removed.``

.github/workflows/staging.yml
nenonja: https://github.com/City-of-Helsinki/palvelutarjotin-admin/blob/07e715c5cea46d3242b42502e738745c312ab2cc/testcafe/utils/settings.ts#L1-L6

src/icons/icons.stories.tsx
nenonja: it seem to work but I will remove it it is by accident

.github/workflows/staging.yml
Sartonon: oh the browser-test:ci uses it, was little bit confusing because env variables are defined at the end of the step

.github/workflows/review.yml
```diff
@@ -34,13 +34,47 @@ jobs:
   review:
     runs-on: ubuntu-latest
     needs: build
-    name: Review
+    name: Review + Acceptance Tests
```
Sartonon, line 37: When are these test "acceptance test" or "smoke tests" or "e2e tests" or "integration tests"? Trying to understand how do we define those?

nenonja: Yes It would be better to have as less synonyms as possible. In this case acceptance tests are same as browser tests. Usually smoke tests are only small subselection of acceptance tests that verify that the core functionality is working in production.

nenonja: https://www.guru99.com/smoke-sanity-testing.html#:~:text=Smoke%20Testing%20verifies%20the%20critical,whereas%20Sanity%20testing%20isn't.

nenonja: Perhaps later when we have more browser tests, some of them are located in different smoke test suite (sub) folder and only them are runned against production. Perhaps we need another yarn script for smoke tests too then.

Sartonon (approved):
Great!

PR #134, opened 2021-01-26
PT-598: Publish event accessibility fix

melniiv:
### Closes :no_good_woman:

### DIFF
--- a/src/domain/event/EventSummaryPage.tsx
+++ b/src/domain/event/EventSummaryPage.tsx
@@ -1,6 +1,5 @@
-import classNames from 'classnames';
 import { isPast } from 'date-fns';
-import { Button, RadioButton } from 'hds-react';
+import { Button } from 'hds-react';
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory, useParams } from 'react-router';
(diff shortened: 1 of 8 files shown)


src/domain/event/EventSummaryPage.tsx
```diff
@@ -146,6 +147,9 @@ const EventSummaryPage: React.FC = () => {
     setShowPublishModal(true);
   };
 
+  //temporary accessibility change pt-598
+  const isAdvancedPublish = false;
```
Sartonon, line 151: Nice way to disable this. Would it be ever nicee to have this outisde of the component as a "constant" like we have this: `const PAST_OCCURRENCE_AMOUNT = 4;`. 

`const IS_ADVANCED_PUBLISH = false;`

Sartonon: This might be also a good place to move that publish section code to its own component (I think it could be in the same file). Just for readability so it is easier to visually see the part that is actually for publishing

Sartonon (commented):
Looking good. Just consider my suggestions :)

src/domain/event/EventSummaryPage.tsx
melniiv: good comments

Sartonon (approved):
Great!

PR #140, opened 2021-02-01
Added studyLevel API changes (based on PT-681)

nikomakela:
THIS is an alternative for https://github.com/City-of-Helsinki/palvelutarjotin-admin/pull/139

I first tried to find a task or some commits that would already have some implementation into this, but I couldn't, so I started from fresh. Now I found this bfe3d91 which is in PT-681-only-one-group-checkbox -branch. I'm going to look how these 2 branches differ from each other - shouldn't much.

There is now yet another branch called feature/PT-729_to_PT-681 that has the same changes, but is rebased to PT-681-only-one-group-checkbox -branch

### DIFF
--- a/src/domain/occurrence/enrolmentDetails/EnrolmentDetails.tsx
+++ b/src/domain/occurrence/enrolmentDetails/EnrolmentDetails.tsx
@@ -20,7 +20,7 @@ import {
   OccurrenceDocument,
   OccurrenceQuery,
   OccurrenceQueryVariables,
-  StudyLevel,
+  StudyLevelNodeConnection,
   useApproveEnrolmentMutation,
   useDeclineEnrolmentMutation,
   useDeleteEnrolmentMutation,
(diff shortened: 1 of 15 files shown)


src/domain/occurrence/enrolmentDetails/EnrolmentDetails.tsx
```diff
@@ -153,25 +153,25 @@ const EnrolmentDetails: React.FC<EnrolmentDetailsProps> = ({
...
-      return studyLevel.map((level) => getStudyLevelString(level)).join(', ');
+          return level.startsWith('GRADE')
+            ? t('studyLevel.grade_interval', {
+                postProcess: 'interval',
+                count: Number(level.split('_')[1]),
+              })
+            : translateValue('studyLevel.', level, t);
+        }
+      );
+      return translations.join(', ');
```
Sartonon, line 171: Common parts with studyLevelOptions in EnrolmentForm.tsx could be maybe extracted to some common helper function 🤔  (yes I noticed that the duplications were there even before your changes :))

nikomakela: A new studyLevel util is now created in commit https://github.com/City-of-Helsinki/palvelutarjotin-admin/commit/ef2bbc9798e6463c7105309798b03d8b5bc8c8f3. Commit is included in pull request !139, which is also currently rebased to this PR's branch, **which makes this PR obsolete!**

PR #142, opened 2021-02-03
PT-718, PT-584 | Focus to top and skip link

Sartonon:
## Description :sparkles:

- Added focus to top functionality on route change
- Added skip link

### Closes :no_good_woman:
**[PT-718](https://helsinkisolutionoffice.atlassian.net/browse/PT-718):** 
**[PT-584](https://helsinkisolutionoffice.atlassian.net/browse/PT-584):** 

## Screenshots :camera_flash:

Screenshot 2021-02-03 at 15 39 40

### DIFF
--- a/src/domain/app/ScrollToTop.tsx
+++ b/src/domain/app/ScrollToTop.tsx
@@ -3,6 +3,8 @@ import { matchPath, useLocation } from 'react-router';
 
 import useLocale from '../../hooks/useLocale';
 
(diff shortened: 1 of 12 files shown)


nenonja (approved):
Code looks good. When I tested however, sometimes when I open an event with lots of content, scrolled to the bottom the click back button, it did not scroll to the top. It only happened only once, so I dont know if it's a problem or just one glitch. Approved, but a thing to consider.

PR #144, opened 2021-02-05
Price field is activated along side with 'is free' -field.

nikomakela:
PT-734 The price field will be active when the free event checkbox is not marked. All the hard coded isFree -values are now fetched from actual data from database.

### DIFF
--- a/src/domain/event/CreateEventPage.tsx
+++ b/src/domain/event/CreateEventPage.tsx
@@ -55,25 +55,6 @@ const CreateEventPage: React.FC = () => {
   };
 
   const handleSubmit = async (values: CreateEventFormFields) => {
-    console.log(
-      JSON.stringify(
-        {
-          variables: {
-            event: {
-              ...getEventPayload({
-                values,
-                selectedLanguage,
-                organisationId: selectedOrganisation?.id || '',
-              }),
-              // save event always as a draft first
-              draft: true,
-            },
-          },
-        },
-        null,
-        2
-      )
-    );
     try {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const requests: Promise<any>[] = [];
--- a/src/domain/event/CreateEventPage.tsx
+++ b/src/domain/event/CreateEventPage.tsx
@@ -55,6 +55,25 @@ const CreateEventPage: React.FC = () => {
   };
 
   const handleSubmit = async (values: CreateEventFormFields) => {
+    console.log(
+      JSON.stringify(
+        {
+          variables: {
+            event: {
+              ...getEventPayload({
+                values,
+                selectedLanguage,
+                organisationId: selectedOrganisation?.id || '',
+              }),
+              // save event always as a draft first
+              draft: true,
+            },
+          },
+        },
+        null,
+        2
+      )
+    );
     try {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const requests: Promise<any>[] = [];
--- a/src/domain/event/__tests__/CreateEventPage.test.tsx
+++ b/src/domain/event/__tests__/CreateEventPage.test.tsx
@@ -100,6 +100,9 @@ const createEventVariables = {
         price: {
           fi: '',
         },
+        description: {
+          fi: '',
+        },
         isFree: true,
       },
     ],
--- a/src/domain/event/__tests__/CreateEventPage.test.tsx
+++ b/src/domain/event/__tests__/CreateEventPage.test.tsx
@@ -731,3 +731,18 @@ const testMultiDropdownValues = async ({
     });
   });
 };
+
+test('price field is accessible only when isFree field is not checked', async () => {
+  render(<CreateEventPage />, { mocks });
+  await waitFor(() => {
+    expect(screen.getByLabelText(/Tapahtuma on ilmainen/)).toBeChecked();
+    expect(screen.getByLabelText(/Hinta/)).toHaveAttribute('disabled');
+    expect(screen.getByLabelText(/Lisätiedot/)).toHaveAttribute('disabled');
+  });
+  userEvent.click(screen.getByLabelText(/Tapahtuma on ilmainen/));
+  await waitFor(() => {
+    expect(screen.getByLabelText(/Tapahtuma on ilmainen/)).not.toBeChecked();
+    expect(screen.getByLabelText(/Hinta/)).not.toHaveAttribute('disabled');
+    expect(screen.getByLabelText(/Lisätiedot/)).not.toHaveAttribute('disabled');
+  });
+});
--- a/src/domain/event/__tests__/CreateEventPage.test.tsx
+++ b/src/domain/event/__tests__/CreateEventPage.test.tsx
@@ -95,7 +95,14 @@ const createEventVariables = {
   event: {
     name: { fi: 'Testitapahtuma' },
     startTime: '2020-08-07T21:00:00.000Z',
-    offers: [{ isFree: true }],
+    offers: [
+      {
+        price: {
+          fi: '',
+        },
+        isFree: true,
+      },
+    ],
     shortDescription: { fi: 'Testikuvaus' },
     description: { fi: 'Pidempi kuvaus' },
     images: [{ internalId: '/image/48584/' }],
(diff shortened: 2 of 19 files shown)


src/domain/event/eventForm/EventForm.tsx
```diff
@@ -395,48 +395,61 @@ const EventForm = <T extends FormFields>({
...
+                            type="number"
+                          />
+                        </FormGroup>
+                      </div>
+                    </div>
                     <div className={styles.priceRow}>
                       <div>
                         <Field
-                          disabled
+                          disabled={isFree === true}
```
Sartonon, line 436: an be simplified -> `disabled={isFree}`

nikomakela:
The API changes should be published before this PR can be used. The paymentInstruction field needs to be removed from API first. Since `event.offers.description` and `event.offers.infoUrl` can be used for same purpose, there is no need for paymentInstruction field and it will be removed as soon as possible.

Sartonon:
Hinta and lisätiedot could use some more spacing between them :)

Screenshot 2021-02-08 at 13 56 25

nikomakela:
> Hinta and lisätiedot could use some more spacing between them :)
> 
> Screenshot 2021-02-08 at 13 56 25

On other fields the bottom space seems to come from a FormGroup -component. In my opinion, all those are the same group - they even depend on each other - so, I don't want to use that component. I need to add something special for it. :)

nikomakela:
> On other fields the bottom space seems to come from a FormGroup -component. In my opinion, all those are the same group - they even depend on each other - so, I don't want to use that component. I need to add something special for it. :)

image

src/domain/event/eventCategorisation/EventCategorisation.tsx
```diff
@@ -21,6 +21,20 @@ const EventCategorisation: React.FC<Props> = ({ eventData, language }) => {
   const inLanguage = eventData.event?.inLanguage;
   const audience = eventData.event?.audience;
   const keywords = eventData.event?.keywords;
+  const offer = eventData.event?.offers?.[0];
+  const price = offer?.price?.[language] || '';
+  const priceDescription = offer?.description?.[language] || '';
+  const isFree = offer?.isFree ?? !price;
+  const priceInfoUrl = offer?.infoUrl?.[language] || '';
```
Sartonon, line 28: Like in the other PR this can be cleaned a tiny bit:

```
  const offer = eventData.event?.offers?.[0];
  const price = offer?.price?.[language];
  const priceDescription = offer?.description?.[language];
  const isFree = offer?.isFree ?? !price;
  const priceInfoUrl = offer?.infoUrl?.[language];
```

just a small nitpick

Sartonon: (default values are not needed)

src/domain/event/eventCategorisation/EventCategorisation.tsx
```diff
@@ -58,7 +72,14 @@ const EventCategorisation: React.FC<Props> = ({ eventData, language }) => {
       <div className={styles.priceRow}>
         <div>
           <TextTitle>{t('eventDetails.categorisation.labelPrice')}</TextTitle>
-          <p>{t('eventDetails.categorisation.eventIsFree')}</p>
+          {isFree && <p>{t('eventDetails.categorisation.eventIsFree')}</p>}
+          {!isFree && (
+            <p className={styles.eventPriceMultiline}>
+              {price}
+              {priceDescription !== '' && `\n${priceDescription}`}
+              {priceInfoUrl !== '' && createLink(`\nInfo: `, priceInfoUrl)}
```
Sartonon, line 80: these can be changed then (little bit cleaner):

```
   {priceDescription && `\n${priceDescription}`}
   {priceInfoUrl && createLink(`\nInfo: `, priceInfoUrl)}
```

src/domain/event/mutation.ts
```diff
@@ -106,6 +114,14 @@ export const MUTATION_EVENT = gql`
...
             altText
           }
+          offers {
+            price {
+              en
+              fi
+              sv
+            }
+            isFree
+          }
```
Sartonon, line 124: offerFields fragment could be used if it doesn't add overhead in the backend?

otherwise I think a new fragment could be created here because this is repeated twice :)

Sartonon (commented):
Looking good! Added couple comments :)

Also problem with error message, not very user friendly :)
Screenshot 2021-02-09 at 10 06 42

Sartonon:
> > Hinta and lisätiedot could use some more spacing between them :)
> 
> On other fields the bottom space seems to come from a FormGroup -component. In my opinion, all those are the same group - they even depend on each other - so, I don't want to use that component. I need to add something special for it. :)

Yeah FormGroup seems to not ake sense in many places, it could be renamed at some point or we could change the way we add spacing to inputs

nikomakela:
API changes are already delivered City-of-Helsinki/palvelutarjotin#122

src/domain/event/mutation.ts
nikomakela: That's true

src/domain/event/eventCategorisation/EventCategorisation.tsx
nikomakela: Changed

src/domain/event/eventCategorisation/EventCategorisation.tsx
nikomakela: Default values removed

src/domain/event/CreateEventPage.tsx
```diff
@@ -55,6 +55,25 @@ const CreateEventPage: React.FC = () => {
...
+              }),
+              // save event always as a draft first
+              draft: true,
+            },
+          },
+        },
+        null,
+        2
+      )
+    );
```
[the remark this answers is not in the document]
nikomakela: Yep! Removed now.

src/domain/event/__tests__/EditEventPage.test.tsx
```diff
@@ -370,3 +382,18 @@ test('returns to create occurrences page when it should after saving', async ()
...
+    expect(screen.getByLabelText(/Tapahtuma on ilmainen/)).toBeChecked();
+    expect(screen.getByLabelText(/Hinta/)).toHaveAttribute('disabled');
+    expect(screen.getByLabelText(/Lisätiedot/)).toHaveAttribute('disabled');
+  });
+  userEvent.click(screen.getByLabelText(/Tapahtuma on ilmainen/));
+  await waitFor(() => {
+    expect(screen.getByLabelText(/Tapahtuma on ilmainen/)).not.toBeChecked();
+    expect(screen.getByLabelText(/Hinta/)).not.toHaveAttribute('disabled');
+    expect(screen.getByLabelText(/Lisätiedot/)).not.toHaveAttribute('disabled');
+  });
```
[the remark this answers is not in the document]
nikomakela: you are absolutely right! Didn't think about it. Now using your code suggestion.

src/domain/event/__tests__/CreateEventPage.test.tsx
```diff
@@ -719,3 +731,18 @@ const testMultiDropdownValues = async ({
...
+    expect(screen.getByLabelText(/Tapahtuma on ilmainen/)).toBeChecked();
+    expect(screen.getByLabelText(/Hinta/)).toHaveAttribute('disabled');
+    expect(screen.getByLabelText(/Lisätiedot/)).toHaveAttribute('disabled');
+  });
+  userEvent.click(screen.getByLabelText(/Tapahtuma on ilmainen/));
+  await waitFor(() => {
+    expect(screen.getByLabelText(/Tapahtuma on ilmainen/)).not.toBeChecked();
+    expect(screen.getByLabelText(/Hinta/)).not.toHaveAttribute('disabled');
+    expect(screen.getByLabelText(/Lisätiedot/)).not.toHaveAttribute('disabled');
+  });
```
Sartonon, line 747: Noticed that this has the same waitFor problem

nikomakela: I converted it to same format as it was in editpage test and removed that edit page test. Needless to test this twice.

Sartonon (approved):

PR #149, opened 2021-02-22
Added adults to a count of approved people.

nikomakela:
PT-782 Fixed approved count that was missed earlier.


src/domain/occurrence/enrolmentTable/EnrolmentTable.tsx
```diff
@@ -82,11 +82,19 @@ const EnrolmentTable: React.FC<Props> = ({
...
+    );
 
   const pendingCount = enrolments
     .filter((e) => e.status === EnrolmentStatus.Pending)
-    .reduce((acc, cur) => acc + cur.studyGroup.groupSize, 0);
+    .reduce(
+      (acc, cur) =>
+        acc + cur.studyGroup.groupSize + cur.studyGroup.amountOfAdult,
+      0
+    );
```
Sartonon, line 97: Could these be combined into a helper function that receives enrolments and EnrolmenStatus as params and returns the count?

nikomakela: Amended! Moved filter-reduce to utils.

Sartonon (approved):
VERY good :)

PR #152, opened 2021-02-26
Added a missing translation for "One group fills the event" -checkbox.

nikomakela:
PT-822 Translations were mixed and 1 was missing.


Sartonon (approved):
Great!

PR #156, opened 2021-03-12
Removed language field from event form

nikomakela:
Removed language field from event form, since the event language is now auto populated by API (https://github.com/City-of-Helsinki/palvelutarjotin/pull/131). **This PR needs that API change in order to work properly!**

PT-877 Added new languages to occurrence language selection.

PT-829 PT-857  Event form's basic information (step 1) included an inLanguage field which is now removed, since it is automatically populated with occurrences languages.

### DIFF
--- a/src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
+++ b/src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
@@ -13,7 +13,8 @@ import FocusToFirstError from '../../../common/components/form/FocusToFirstError
 import FormGroup from '../../../common/components/form/FormGroup';
 import FormHelperText from '../../../common/components/FormHelperText/FormHelperText';
 import TextTitle from '../../../common/components/textTitle/TextTitle';
-import { EventQuery, Language } from '../../../generated/graphql';
+import { EVENT_LANGUAGES } from '../../../constants';
+import { EventQuery } from '../../../generated/graphql';
 import useLocale from '../../../hooks/useLocale';
 import formatDate from '../../../utils/formatDate';
 import PlaceInfo from '../../place/placeInfo/PlaceInfo';
@@ -187,7 +188,7 @@ const EventOccurrenceForm: React.FC<Props & GoToPublishingProps> = ({
                     label={t('eventOccurrenceForm.labelLanguages')}
                     name="languages"
                     options={[
-                      ...Object.values(Language).map((language) => ({
+                      ...Object.values(EVENT_LANGUAGES).map((language) => ({
                         label: t(`common.languages.${language.toLowerCase()}`),
                         value: language,
                       })),
--- a/src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
+++ b/src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
@@ -17,6 +17,7 @@ import { EVENT_LANGUAGES } from '../../../constants';
 import { EventQuery } from '../../../generated/graphql';
 import useLocale from '../../../hooks/useLocale';
 import formatDate from '../../../utils/formatDate';
+import sortFavorably from '../../../utils/sortFavorably';
 import PlaceInfo from '../../place/placeInfo/PlaceInfo';
 import VenueDataFields from '../../venue/venueDataFields/VenueDataFields';
 import { OccurrenceFormFields } from '../types';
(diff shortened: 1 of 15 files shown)


src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
```diff
@@ -87,6 +89,31 @@ const EventOccurrenceForm: React.FC<Props & GoToPublishingProps> = ({
...
+        EVENT_LANGUAGES.EN,
+      ].map((language) => t(`common.languages.${language.toLowerCase()}`))
+    );
+
+    const languages = [
+      ...Object.values(EVENT_LANGUAGES).map((language) => ({
+        label: t(`common.languages.${language.toLowerCase()}`),
+        value: language,
+      })),
+    ].sort(
```
Sartonon, line 109: Creating new array and spreading is redundant here, this is enough:

```
 const languages = Object.values(EVENT_LANGUAGES)
      .map((language) => ({
        label: t(`common.languages.${language.toLowerCase()}`),
        value: language,
      }))
      .sort(
        (a, b) =>
          languagesOrder.indexOf(a.label) - languagesOrder.indexOf(b.label)
      );
```

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
```diff
@@ -87,6 +89,31 @@ const EventOccurrenceForm: React.FC<Props & GoToPublishingProps> = ({
...
+      ),
+      [
+        EVENT_LANGUAGES.FI,
+        EVENT_LANGUAGES.SV,
+        EVENT_LANGUAGES.EN,
+      ].map((language) => t(`common.languages.${language.toLowerCase()}`))
+    );
+
+    const languages = [
+      ...Object.values(EVENT_LANGUAGES).map((language) => ({
```
Sartonon, line 105: Can you use languagesOrder here so that we wouldn't need another sort?

nikomakela: The problem is that `languagesOrder` does not contain the value - it's a list of labels. I tried to create an alternative for `sortFavorably` that would have sorted objects with a given key, but that would have been a mess in TypeScript. Since it's needed only here, I thought it would be better to do it like this.

The thing with the complex sorter is that there are 6 languages total, but we want to add 3 of them (fi, sv, en) in the beginning of the array and rest should be sorted alphabetically (..., ar, ru, zh).

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
nikomakela: Fixed.

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
```diff
@@ -87,6 +89,31 @@ const EventOccurrenceForm: React.FC<Props & GoToPublishingProps> = ({
...
+        label: t(`common.languages.${language.toLowerCase()}`),
+        value: language,
+      })),
+    ].sort(
+      (a, b) =>
+        languagesOrder.indexOf(a.label) - languagesOrder.indexOf(b.label)
+    );
+
+    return languages;
+  };
```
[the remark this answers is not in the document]
nikomakela: Added it. `t` was automatically set as a dependency.

nikomakela:
Strange that this passes acceptance tests against wrong version of API. Structure of the language query has changed.

src/utils/sortFavorably.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Sort favored items in the beginning of the list in given order.
+ * Then sort the rest of the items alphabetically.
+ */
+export default function sortFavorably(sortable: string[], favored: string[]) {
+  favored.reverse();
```
Sartonon, line 6: Would better to not mutate the array given as parameter to avoid unwanted side-effects.

```
favored.slice().reverse();
```

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
Sartonon: Just thinking if that would just work correctly? Or is it wrong order, can't test it at the moment :D

```
languagesOrder.map((language) => ({
        label: t(`common.languages.${language.toLowerCase()}`),
        value: language,
      }))
```

Sartonon (approved):

src/utils/sortFavorably.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Sort favored items in the beginning of the list in given order.
+ * Then sort the rest of the items alphabetically.
+ */
+export default function sortFavorably(sortable: string[], favored: string[]) {
+  favored = [...favored].reverse();
+  return [...sortable].sort((a: string, b: string) => {
```
Sartonon, line 7: No need to use spread here again, just sortable.sort...

src/utils/sortFavorably.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Sort favored items in the beginning of the list in given order.
+ * Then sort the rest of the items alphabetically.
+ */
+export default function sortFavorably(sortable: string[], favored: string[]) {
+  favored = [...favored].reverse();
```
Sartonon, line 6: btw is it necessary to reverse  this in the first place, could handle ifs differently or is there some other idea behind this?

nikomakela: without reverse "the rest of the languages" would be in the beginning and favored would be in the end of the returned array. I don't know why, but it just is so :D 🤯

PR #158, opened 2021-03-17
Improved occurrence form's date field validation

nikomakela:
PT-481 Occurrence's date is validated against pEvent.enrolmentStart and pEvent.enrolmentEndDays fields.

Submitted as a draft, since this branch is basing on PR #156. It's still a ready implementation.

It would be nice to get improvement suggestions for passing the 2 new pEvent variables to occurrence validation. Now they are passed via occurrence form model.

### DIFF
--- a/src/domain/occurrence/CreateOccurrencePage.tsx
+++ b/src/domain/occurrence/CreateOccurrencePage.tsx
@@ -11,7 +11,6 @@ import BackButton from '../../common/components/backButton/BackButton';
 import EventSteps from '../../common/components/EventSteps/EventSteps';
 import LoadingSpinner from '../../common/components/loadingSpinner/LoadingSpinner';
 import {
-  EventQuery,
   OccurrenceFieldsFragment,
   useAddOccurrenceMutation,
   useDeleteOccurrenceMutation,
@@ -75,7 +74,7 @@ const CreateOccurrencePage: React.FC = () => {
     (item) => !isPast(new Date(item.startTime))
   );
 
-  const initialFormValues = useInitialFormValues(isFirstOccurrence, eventData);
+  const initialFormValues = useInitialFormValues(isFirstOccurrence);
   const { loading: loadingMyProfile } = useMyProfileQuery();
 
   const [createOccurrence] = useAddOccurrenceMutation();
@@ -197,7 +196,7 @@ const CreateOccurrencePage: React.FC = () => {
         isLoading={eventIsInitialLoading || loadingMyProfile}
         hasPadding={false}
       >
-        {eventData ? (
+        {eventData?.event ? (
           <>
             {isEditableEvent(eventData) ? (
               <Container>
@@ -223,7 +222,7 @@ const CreateOccurrencePage: React.FC = () => {
                     />
                   )}
                   <EventOccurrenceForm
-                    eventData={eventData}
+                    event={eventData.event}
                     formTitle={t('createOccurrence.formTitle')}
                     initialValues={initialFormValues}
                     onCancel={goToEventSummary}
@@ -253,10 +252,7 @@ const CreateOccurrencePage: React.FC = () => {
 
 // TODO: maybe could just provide enableReinitialize props form parent component
 // to simplify this. We might not need reinitialization here.
-const useInitialFormValues = (
-  isFirstOccurrence: boolean,
-  eventData: EventQuery | undefined
-) => {
+const useInitialFormValues = (isFirstOccurrence: boolean) => {
   const searcParams = useSearchParams();
 
   const getInitialFormValues = () => {
@@ -265,36 +261,27 @@ const useInitialFormValues = (
     const initialStartsAt = searcParams.get('startsAt');
     const initialEndsAt = searcParams.get('endsAt');
 
-    const initialValues = {
-      ...defaultInitialValues,
-      enrolmentStart: eventData?.event?.pEvent?.enrolmentStart
-        ? new Date(eventData?.event?.pEvent?.enrolmentStart)
-        : null,
-      enrolmentEndDays: eventData?.event?.pEvent?.enrolmentEndDays || 0,
-    };
-
     // if is first occurrence, use pre-filled values from event form (query params)
     if (isFirstOccurrence && initialDate && initialEndsAt && initialStartsAt) {
-      const valuesAreValid =
+      if (
         isValidDate(new Date(initialDate)) &&
         isValidTime(initialStartsAt) &&
-        isValidTime(initialEndsAt);
-
-      if (valuesAreValid) {
-        Object.assign(initialValues, {
+        isValidTime(initialEndsAt)
+      ) {
+        return {
+          ...defaultInitialValues,
           date: new Date(initialDate),
           startsAt: initialStartsAt,
           endsAt: initialEndsAt,
-        });
+        };
       }
     }
-
-    return initialValues;
+    return defaultInitialValues;
   };
 
   // we don't want the initialValues to update because we don't
   // want form to reset to those on subsequent renders
-  return React.useMemo(getInitialFormValues, [eventData]);
+  return React.useMemo(getInitialFormValues, []);
 };
 
 export default CreateOccurrencePage;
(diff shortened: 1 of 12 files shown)


src/domain/occurrence/CreateOccurrencePage.tsx
```diff
@@ -261,27 +265,38 @@ const useInitialFormValues = (isFirstOccurrence: boolean) => {
...
-          }
-        : defaultInitialValues;
+
+      if (valuesAreValid) {
+        Object.assign(initialValues, {
+          date: new Date(initialDate),
+          startsAt: initialStartsAt,
+          endsAt: initialEndsAt,
+        });
+      }
```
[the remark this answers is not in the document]
nikomakela: It wasn't until I added 2 new conditional fields there 🦹‍♂️

src/domain/occurrence/CreateOccurrencePage.tsx
```diff
@@ -261,27 +265,38 @@ const useInitialFormValues = (isFirstOccurrence: boolean) => {
...
     const initialEndsAt = searcParams.get('endsAt');
 
+    const initialValues = Object.assign(
+      { ...defaultInitialValues },
+      {
+        enrolmentStart: eventData?.event?.pEvent?.enrolmentStart
+          ? new Date(eventData?.event?.pEvent?.enrolmentStart)
+          : null,
+        enrolmentEndDays: eventData?.event?.pEvent?.enrolmentEndDays || 0,
+      }
```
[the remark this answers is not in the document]
nikomakela: Okay, I need to check that. With that solution I could manage without so many changes. I could make an alternative PR or something, let's see.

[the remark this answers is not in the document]
nikomakela: After rethinking this, what would be the purpose of `useInitialFormValues`, if it isn't to give all the values needed by the form. I understood that on-the-fly created validation scheme that you suggested @Sartonon  would have been a solution for not to include eventData in useInitialFormValues? And that is the thing that also I requested and first thought that would be a good thing, but now I've changed my mind, because if the validation schema is created on the fly from multiple pieces, it would only make it hard to follow, what really is validated. Now everything have their place.

Or did you mean something else with this suggestion @Sartonon?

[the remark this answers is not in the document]
Sartonon: Its purpose would be to provide initialValues from the event form for the first occurrence?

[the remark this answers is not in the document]
Sartonon: Current solution just feels a little bit hacky, we are providing some initialValues to form that are only used for validation. It seems like enrolmentStart and enrolmentEndDays are included in the form. I remember being very confused in some other form where I saw some initial values but couldn't find the in the form and it is not very easy to see that they are only used in validation.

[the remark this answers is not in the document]
Sartonon: Someone could easily look at the forms initialValues and just delete some of them because he/she would think that maybe those were used previously in the form but had been removed but doesnt realise that they are used in the validationSchema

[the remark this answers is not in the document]
nikomakela: Those 2 event fields are no longer part of occurrence form. I updated the validationSchema to be create on the fly before Formik, but I'm not sure if it is what you meant.

src/domain/occurrence/eventOccurrenceForm/ValidationSchema.ts
```diff
@@ -95,3 +98,38 @@ export default Yup.object().shape({
...
+    const minDate =
+      enrolmentEndDays > 0
+        ? addDays(enrolmentStart, enrolmentEndDays)
+        : enrolmentStart;
+    minDate.setHours(0, 0, 0, 0);
+    return schema.min(minDate, () => ({
+      key: VALIDATION_MESSAGE_KEYS.DATE_MIN,
+      min: formatDate(minDate),
+    }));
+  }
```
Sartonon, line 127: I would move this function one outside of this function because all parameters are passed anyway. To that would be easier to read

src/domain/occurrence/CreateOccurrencePage.tsx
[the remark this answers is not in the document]
Sartonon: I think you did exactly as I described :)

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
```diff
@@ -112,6 +115,16 @@ const EventOccurrenceForm: React.FC<Props & GoToPublishingProps> = ({
       );
   }, [t]);
 
+  const validationSchema = React.useMemo(
+    () =>
+      testDateWithEnrolmentValidationSchema(
+        ValidationSchema.fields.date as DateSchema,
+        event?.pEvent,
+        ValidationSchema
```
Sartonon, line 123: Hmm is there any point passing validationSchema as a parameter? Would it be better to just use it in the function, they are in the same file?

nikomakela: I agree there isn't, but I just extend the default validation schema as hard coded, I don't see there much difference to the first version where the date field dependencies were part of the default schema. We use this schema only once and then the dependencies are always needed. Now I have just splitted it to 2 schemas where 1 is extending the another and only the extended one is ever used.

src/domain/occurrence/eventOccurrenceForm/ValidationSchema.ts
nikomakela: Moved.

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
Sartonon: If the extended one is only used the there shouldn't be any need fo extension :)

src/domain/occurrence/eventOccurrenceForm/ValidationSchema.ts
```diff
@@ -95,3 +98,41 @@ export default Yup.object().shape({
...
+      ? addDays(enrolmentStart, enrolmentEndDays)
+      : enrolmentStart;
+  minDate.setHours(0, 0, 0, 0);
+  return schema.min(minDate, () => ({
+    key: VALIDATION_MESSAGE_KEYS.DATE_MIN,
+    min: formatDate(minDate),
+  }));
+};
+
+export const testDateWithEnrolmentValidationSchema = (
```
Sartonon, line 118: I feel like this could be just named `getValidationSchema`. Maybe there could be comment in the function to clarify the validation logic

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
Sartonon: or is there any chance that either  enrolmentStart or  enrolmentEndDays is null/undefined?

nikomakela: > If the extended one is only used the there shouldn't be any need fo extension :)

Well, the first version of PR didn't have one...

> or is there any chance that either enrolmentStart or enrolmentEndDays is null/undefined?

they should always be filled, because every occurrence needs an event and they are mandatory properties for an event.

src/domain/occurrence/eventOccurrenceForm/ValidationSchema.ts
nikomakela: changed and commented.

src/domain/occurrence/eventOccurrenceForm/EventOccurrenceForm.tsx
Sartonon: > Well, the first version of PR didn't have one...
The problem was that we have values in forms initialValues that are not actually in the form but only used in validation which I fell is little bit confusing but maybe in the end it is matter of taste

Sartonon (approved):

PR #160, opened 2021-03-22
Some new help texts to event form.

nikomakela:
PT-804 Added a guidance text for additional information mandatority. 
PT-811 Added placeholder text for an additional information for payment field.

### DIFF
--- a/src/domain/app/i18n/en.json
+++ b/src/domain/app/i18n/en.json
@@ -286,6 +286,7 @@
       "imageAltTextHelp": "Description of the contents of the image. Add license information also.",
       "labelDescription": "Description",
       "labelMandatoryAdditionalInformation": "Entering additional information is mandatory for the registrant",
+      "guidanceTextMandatoryAdditionalInformation": "Tell above if you need more information about group or other things in activities and make answer as mandatory...",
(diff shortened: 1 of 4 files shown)


src/domain/app/i18n/en.json
```diff
@@ -286,6 +286,7 @@
       "imageAltTextHelp": "Description of the contents of the image. Add license information also.",
       "labelDescription": "Description",
       "labelMandatoryAdditionalInformation": "Entering additional information is mandatory for the registrant",
+      "guidanceTextMandatoryAdditionalInformation": "Tell above if you need more information about group or other things in activities and make answer as mandatory...",
```
Sartonon, line 289: rather than using guidanceText, helperText would be more in line with other translations and the prop name

nikomakela: I selected guidance text only because it's not an actual helperText. Checkboxes seems to not have them. Also that text is related to 2 fields. But I don't mind the name, I can change it if needed.

Sartonon: OK, no need to change :)

Sartonon (approved):

PR #161, opened 2021-03-23
Use common language translations everywhere.

nikomakela:
PT-908. Using only 1 common language translation instead of 3 different ones, which content is still the same. Helps maintaining the code and the translations.  Also added  a new component named OccurrenceGroupLanguageInfo, which is having the same behavior that the component in palvelutarjotin-ui repo.

image

image

image


Sartonon (approved):
Looks good!

PR #166, opened 2021-04-09
Added a missing translation for occurrence section help text.

nikomakela:
PT-843.


Sartonon (approved):
Good. Unfortunately this will be deleted in the redesign :)

PR #177, opened 2021-05-07
Allow CIMODE as a language and list it in a language selector.

nikomakela:
PT-1015. CIMODE is now allowed by the router and it is selectable from a list of languages if the feature flag is set on.

PT-1015. useLocale cannot be used for history pushes, because it does not handle the cimode. If the cimode would be added to useLocales languages, all the datepickers, autosuggest fields, etc would stop working, because they don't support the cimode at all. To fix the issue, router needs to be controlled with i18next instances language instead of useLocal hook.

image


nikomakela:
I'm not happy of that I needed to change the history pushes locale set. useLocale would be a really great hook to use everywhere as it was earlier, but the problem is that I couldn't find a way to make it support the cimode. Listing cimode there, would break all the "widgets".

nikomakela:
Feature flag did not work in Teachers UI (not in review, not in staging). It seems not to work here in providers UIs review server either.

src/domain/app/header/languageDropdown/LanguageDropdown.tsx
```diff
@@ -6,26 +6,31 @@ import { useHistory, useLocation } from 'react-router-dom';
...
+import { LanguageSelectorLanguage } from '../../../../types';
 import updateLocaleParam from '../../../../utils/updateLocaleParam';
 
 const LanguageDropdown: React.FC = () => {
-  const { t } = useTranslation();
+  const { t, i18n } = useTranslation();
   const history = useHistory();
   const { pathname, search } = useLocation();
-  const locale = useLocale();
+  const locale = i18n.language;
```
Sartonon, line 17: Could useLocale just be updated to support 'cimode'?

.env.development.local.example
```diff
@@ -5,4 +5,5 @@ REACT_APP_API_URI=http://localhost:8081/graphql
 REACT_APP_LINKEDEVENTS_API_URI=https://api.hel.fi/linkedevents-test/v1
 
 REACT_APP_SENTRY_DSN=
-REACT_APP_ENVIRONMENT=local
\ No newline at end of file
+REACT_APP_ENVIRONMENT=local
+REACT_APP_LANGUAGE_CIMODE_VISIBLE=true
```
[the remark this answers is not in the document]
nikomakela: It should be there 
```
# CIMODE Language button visibility
ARG REACT_APP_LANGUAGE_CIMODE_VISIBLE
```

src/domain/app/header/languageDropdown/LanguageDropdown.tsx
nikomakela: Nope, I don't think so, like I said in earlier comment: https://github.com/City-of-Helsinki/palvelutarjotin-admin/pull/177#issuecomment-834150214. When the 'cimode' is added there, the auto-suggest -widget, all kind of datepicker widgets, etc starts complaining about it, because they don't support 'cimode' as a language.

src/domain/app/header/__tests__/Header.test.tsx
```diff
@@ -30,3 +31,28 @@ it('focuses skip link first', () => {
...
+    process.env.REACT_APP_LANGUAGE_CIMODE_VISIBLE = 'true';
+    const { container } = render(<Header />);
+    await findAllByText(container, /CIMODE/);
+  });
+
+  test('cimode is not rendered in a list of languages when feature flag is off', async () => {
+    process.env.REACT_APP_LANGUAGE_CIMODE_VISIBLE = 'false';
+    const { container } = render(<Header />);
+    expect(queryAllByText(container, /CIMODE/).length).toBe(0);
+  });
```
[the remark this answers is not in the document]
nikomakela: changed.

src/domain/app/header/languageDropdown/LanguageDropdown.tsx
Sartonon: yeah I see. That is problematic. Maybe history.push should have some abstraction that adds the locale behind the scenes and we don't have to mabually worry about it everhywhere. That would simplify things in this case too.

nikomakela:
Rebased to master.

nikomakela:
1 test fails here in Github, but not on my local machine =/ It needs some investigation

nikomakela:
> 1 test fails here in Github, but not on my local machine =/ It needs some investigation

It seems that it can fail in local development env also, like once with 5 tries.

PR #178, opened 2021-05-08
PT-1013 | Add instruction text for newly signed providers

Sartonon:
## Description :sparkles:
- Add notification for newly signed users

### Closes :no_good_woman:
**[PT-1013](https://helsinkisolutionoffice.atlassian.net/browse/PT-1013):** 

## Screenshots :camera_flash:

Screenshot 2021-05-08 at 11 28 46
Screenshot 2021-05-08 at 11 28 36

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+graphql.ts 
+graphql.tsx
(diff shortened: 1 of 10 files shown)


nikomakela (approved):
The changes are looking good. Now when we have a way to check if the user has write permissions, should the create event -button be hidden, when there are no write-permissions? That could be done in a new task.

Sartonon:
> The changes are looking good. Now when we have a way to check if the user has write permissions, should the create event -button be hidden, when there are no write-permissions? That could be done in a new task.

Yeah true didn't think of that

PR #185, opened 2021-05-13
Added download events' enrolments button to event summary page.

nikomakela:
PT-1024 PT-1032. Added apiReportClient which is implemented with Axios. When the API does not return csv file, toast is thrown. Events' enrolments csv file can be downloaded when API does not require authentication.

### DIFF
--- a/src/clients/apiReportClient/apiReportClient.ts
+++ b/src/clients/apiReportClient/apiReportClient.ts
@@ -1,23 +1,23 @@
 import axios from 'axios';
 
-// import { store } from '../../domain/app/store';
-// import { apiTokenSelector } from '../../domain/auth/selectors';
+import { store } from '../../domain/app/store';
+import { apiTokenSelector } from '../../domain/auth/selectors';
 
 const axiosClient = axios.create({
   baseURL: `${process.env.REACT_APP_API_REPORT_URI}`,
   timeout: 30000,
 });
 
-// axiosClient.interceptors.request.use(
-//   (config) => {
-//     const token = apiTokenSelector(store.getState());
-//     if (token) {
-//       config.headers.Authorization = `Bearer ${token}`;
-//     }
-//     return config;
-//   },
-//   (error) => Promise.reject(error)
-// );
+axiosClient.interceptors.request.use(
+  (config) => {
+    const token = apiTokenSelector(store.getState());
+    if (token) {
+      config.headers.authorization = `Bearer ${token}`;
+    }
+    return config;
+  },
+  (error) => Promise.reject(error)
+);
 
 export enum ROUTES {
   EVENTS_ENROLMENTS = '/palvelutarjotinevent/enrolments/csv/',
--- a/src/clients/apiReportClient/apiReportClient.ts
+++ b/src/clients/apiReportClient/apiReportClient.ts
@@ -1,0 +1,26 @@
+import axios from 'axios';
+
+// import { store } from '../../domain/app/store';
+// import { apiTokenSelector } from '../../domain/auth/selectors';
+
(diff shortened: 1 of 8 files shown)


nikomakela:
The pull request is started as a drafter, because there is still a problem with authentication / authorization. I thought that I could use the same Bearer token that is used with graphQL queries, but I haven't managed to implement such a feature in Django view yet.

src/clients/apiReportClient/useReportClientQuery.ts
```diff
@@ -0,0 +1,40 @@
...
+      document.body.appendChild(link);
+      link.click();
+      link.remove();
+    } else {
+      toast(t('eventEnrolmentsReport.downloadError'), {
+        type: toast.TYPE.ERROR,
+      });
+    }
+  };
+}
```
Sartonon, line 38: Trying to think if creating custom hook for this purpose is benefical. Maybe usage of translation hook justifies this but also function could throw error if status !== 200 and toast be handled in the component.

src/clients/apiReportClient/apiReportClient.ts
```diff
@@ -0,0 +1,26 @@
...
+//     return config;
+//   },
+//   (error) => Promise.reject(error)
+// );
+
+export enum ROUTES {
+  EVENTS_ENROLMENTS = '/palvelutarjotinevent/enrolments/csv/',
+}
+
+export default axiosClient;
```
Sartonon, line 26: Would it make sense to export helper functions that use axiosClient? Rathen than passing ROUTE.EVENTS_ENROLMENTS in other file

src/clients/apiReportClient/useReportClientQuery.ts
```diff
@@ -0,0 +1,40 @@
...
+    // Hack to download files with axios:
+    // https://gist.github.com/javilobo8/097c30a233786be52070986d8cdb1743
+    if (response.status === 200 && response?.data?.type === 'text/csv') {
+      const url = window.URL.createObjectURL(new Blob([response.data]));
+      const link = document.createElement('a');
+      link.href = url;
+      link.setAttribute('download', filename);
+      document.body.appendChild(link);
+      link.click();
+      link.remove();
```
Sartonon, line 31: Could be extracted to helper function called `downloadFile(data)`

src/clients/apiReportClient/useReportClientQuery.ts
nikomakela: I'm quite sure we will have that range query as well, e.g spring 2021 or autumn 2020, etc. That's query's option, not components. Does that make it more beneficial? Sure I could still create a new button-component that downloads this export. @Sartonon what you think?

src/clients/apiReportClient/apiReportClient.ts
nikomakela: I would actually like to tie the routes more to the client. Url mappings should be one of the clients features. But this also depends a lot on what we decide to do with the hook approach. I thought we could have a for every request, but if we don't use hooks, a helper library that uses client's features would be an alternative.

src/clients/apiReportClient/useReportClientQuery.ts
nikomakela: True, It's so hacky hack that it should be isolated as much as possible :)

src/clients/apiReportClient/useReportClientQuery.ts
Sartonon: To me it is ok to with the approach you think is the best.

But I'm not sure if you understood what I meant :)

Sartonon (approved):

PR #199, opened 2021-07-14
Submit occurrence form if needed when saving enrolment form data

Sartonon:
## Description :sparkles:

- Submit also occurrence form when using save button for enrolment info form
- Also submit occurrence form when navigating to publish page if occurrence form is valid and filled

- Disable submit button on main event form when clicked and requests are pending

### Closes :no_good_woman:
**[PT-1047](https://helsinkisolutionoffice.atlassian.net/browse/PT-1047):** 


nikomakela (approved):
💎

PR #204, opened 2021-08-18
fix: hard coded a Finnish registration pending text instead of lorem ipsum.

nikomakela:
In future, instead of using a hard coded (Finnish only) text and a contact information address-block, the content should be fetched from to-be-done headless CMS.

PT-1113.


Sartonon (approved):
Looking good!

PR #206, opened 2021-08-19
PT-1131 | fix pagination when server returns error

Sartonon:
## Description :sparkles:

- Fix bug where event list would disappear after pagination query return error from server

### Closes :no_good_woman:
**[PT-1131](https://helsinkisolutionoffice.atlassian.net/browse/PT-1131):** 


nikomakela (approved):
ok!

PR #214, opened 2021-09-03
feat(profile): added a language selector to my profile form.

nikomakela:
New language input can be used to determine which language will be used for communication.

PT-1147.

image

image


Sartonon (approved):
Looks Good

PR #218, opened 2021-09-15
feat(enrolment-type): events enrolment info differs on different enrolment type

nikomakela:
The enrolment information is shown only If the event's enrolment type is internal. If the enrolments are done externally or the event is unenrollable, enrolment status info won't be rendered. In event card also shows the enrolment type.

PT-849 PT-1102.

### DIFF
--- a/src/domain/event/EventSummaryPage.tsx
+++ b/src/domain/event/EventSummaryPage.tsx
@@ -50,7 +50,6 @@ const EventSummaryPage: React.FC = () => {
     loading,
     refetch: refetchEventData,
   } = useEventQuery({
-    // fetchPolicy: 'network-only',
     variables: {
       id: eventId,
       include: ['location', 'keywords', 'in_language'],
@@ -62,9 +61,11 @@ const EventSummaryPage: React.FC = () => {
   );
   const [cancelOccurrence] = useCancelOccurrenceMutation();
 
+  const enrolmentType = eventData?.event && getEnrolmentType(eventData.event);
   const organisationId = eventData?.event?.pEvent?.organisation?.id || '';
   const isEventDraft =
     eventData?.event?.publicationStatus === PUBLICATION_STATUS.DRAFT;
+  const isInternalEnrolment = enrolmentType === EnrolmentType.Internal;
 
   const occurrences =
     (eventData?.event?.pEvent?.occurrences.edges.map(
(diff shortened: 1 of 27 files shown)


src/domain/event/EventSummaryPage.tsx
```diff
@@ -189,7 +193,7 @@ const EventSummaryPage: React.FC = () => {
                       })}
                     </span>
                   </h2>
-                  {!isEventDraft && (
+                  {!isEventDraft && enrolmentType === EnrolmentType.Internal && (
```
Sartonon, line 196: minor but result of `enrolmentType === EnrolmentType.Internal` could be saved to a variable just like isEventDraft to make this little bit more concise and easier to read

src/domain/event/eventCard/EventCard.tsx
```diff
@@ -52,6 +55,18 @@ const EventCard: React.FC<Props> = ({
...
+    if (enrolmentType === EnrolmentType.Internal) {
+      return t('events.eventCard.textEnrolments', {
+        count: enrolmentsCount || 0,
+      });
+    } else if (enrolmentType === EnrolmentType.External) {
+      return t('events.eventCard.textExternalEnrolments');
+    } else if (enrolmentType === EnrolmentType.Unenrollable) {
+      return t('events.eventCard.textUnenrollable');
+    }
+  };
```
Sartonon, line 68: Maybe object mapper could be used here to avoid `else if`

Sartonon: or switch statement

src/domain/occurrences/occurrencesTableReadOnly/OccurrencesTableSummary.tsx
```diff
@@ -44,6 +46,64 @@ const OccurrencesTableSummary: React.FC<Props> = ({
     );
   };
 
+  const enrolmentType = eventData?.event && getEnrolmentType(eventData.event);
+
+  const enrolmentColumns =
+    EnrolmentType.Internal === enrolmentType
+      ? [
```
Sartonon, line 53: To me this seems little bit nasty to use ternary here :D 

how about:

```
let enrolmentColumns = [];
if (EnrolmentType.Internal === enrolmentType) {
  enrolmentColumns = [.....]
}
```

Sartonon: or if this doesn't break things:

```
  },
    ...(isInternalEnrolment ? enrolmentColumns : []),
    {
```

nikomakela:
Test results: 

It was said that the event does not have enrolments

image

When the occurrences edit form is loaded, it is initialized with internal enrolments (and empty fields)

image

PR #223, opened 2021-09-23
fix: hide min and max fields when not internal enrolment + improve tests

Sartonon:
## Description :sparkles:

- Hide min and max fields when event doesn't have internal enrolment

### Closes :no_good_woman:
**[PT-1215](https://helsinkisolutionoffice.atlassian.net/browse/PT-1215):** 


nikomakela (approved):
gj

PR #224, opened 2021-09-23
fix: hide also seats input when not internal enrolment

Sartonon:
## Description :sparkles:

- Hide also seats input when not internal enrolments

### Closes :no_good_woman:
**[PT-1215](https://helsinkisolutionoffice.atlassian.net/browse/PT-1215):** 


nikomakela (approved):
GJ

PR #229, opened 2021-10-19
PT-1220 | Fix how multiday occurrences are shown

Sartonon:
Fix how multiday occurrences are shown and update date formatting to comply with HDS guidelines.

### Closes :no_good_woman:
**[PT-1220](https://helsinkisolutionoffice.atlassian.net/browse/PT-1220):** 

## Screenshots :camera_flash:

OLD:

Screenshot 2021-10-19 at 16 18 49
Screenshot 2021-10-19 at 16 19 06

NEW:
Screenshot 2021-10-19 at 16 19 30
Screenshot 2021-10-19 at 16 19 37
Screenshot 2021-10-19 at 16 19 44

Event card updated also:
Screenshot 2021-10-25 at 14 55 29

### DIFF
--- a/src/domain/event/eventPreviewCard/EventPreviewCard.tsx
+++ b/src/domain/event/eventPreviewCard/EventPreviewCard.tsx
@@ -12,7 +12,7 @@ import EventKeywords from '../eventKeywords/EventKeywords';
 import {
   getEventFields,
   getEventPlaceholderImage,
-  getEventStartTimeStr,
+  getNextOccurrenceDateStr,
 } from '../utils';
 import styles from './eventPreviewCard.module.scss';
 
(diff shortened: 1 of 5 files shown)


src/domain/occurrence/__tests__/OccurrenceDetailsPage.test.tsx
```diff
@@ -1,3 +1,5 @@
+import { MockedResponse } from '@apollo/client/testing';
+import { axe } from 'jest-axe';
```
Jokinen, line 2: Looks like this import isn't used.

Jokinen (approved):
The changes looked good to me, although it's not easy to spot the details as there are a lot fo similar changes.

Seems like it could make sense to either move some of the utils into HDS or into https://github.com/City-of-Helsinki/helsinki-utils-js although I'm not sure if we have the resources to do that. Now that the formats should be shared, there's less chance for divergence between projects, which makes sharing easier.

PR #233, opened 2021-11-05
feat: updated the study group and servicemap api schemas.

nikomakela:
Updated the schema with minimal changes: Unit Id and new unit information should be used in UI. Occurrences were somehow set to non-nullable, but they should be nullable. PT-1264 PT-1292.

Dependencies: https://github.com/City-of-Helsinki/palvelutarjotin/pull/221 and https://github.com/City-of-Helsinki/palvelutarjotin/pull/220


src/domain/event/EventSummaryPage.tsx
```diff
@@ -73,7 +73,7 @@ const EventSummaryPage: React.FC = () => {
   const isInternalEnrolment = enrolmentType === EnrolmentType.Internal;
 
   const occurrences =
-    (eventData?.event?.pEvent?.occurrences.edges.map(
+    (eventData?.event?.pEvent?.occurrences?.edges.map(
```
nikomakela, line 76: I don't know when this has changed. I thought it has always been so that occurrences can be null.

PR #236, opened 2021-11-11
feat: unit place selector uses schoolsAndKindergartens query

nikomakela:
API returns all the units (1000+ pcs) all at once from schoolsAndKindergartens graphql query. The Enrolment form autosuggest field for unit selection should be able to filter the list of units with an input value.

PT-1298 PT-1313.

Same ticket in palvelutarjotin-ui: https://github.com/City-of-Helsinki/palvelutarjotin-ui/pull/173

----

The list of schools and kindergartens is from Servicemap API, but the selected place is queried from LinkedEvents places-API

image

with error:

image

### DIFF
--- a/src/domain/place/placeText/PlaceText.tsx
+++ b/src/domain/place/placeText/PlaceText.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { useTranslation } from 'react-i18next';
 
 import { usePlaceQuery } from '../../../generated/graphql';
 import useLocale from '../../../hooks/useLocale';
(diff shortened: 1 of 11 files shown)


nikomakela:
- [ ] Tests are still missing

src/domain/place/placeText/PlaceText.tsx
```diff
@@ -1,20 +1,29 @@
...
   const { data } = usePlaceQuery({
     variables: { id: placeId },
   });
 
-  return <>{getLocalisedString(data?.place?.name || {}, locale)}</>;
+  const text = getLocalisedString(data?.place?.name || {}, locale);
+  const errorText =
+    showErrorText && placeId && !text
+      ? t('common.placeText.noPlaceFoundError')
+      : '';
```
Sartonon, line 24: Same as in kultus-ui PR. I think this doesn't belong in this component

Sartonon (changes requested):
Looks good but some code should be moved out of PlaceText component :)

Sartonon (approved):
Looks good :)

PR #240, opened 2021-11-29
PT-1353 | Add is orderable to school checkbox

Sartonon:
## Description :sparkles:

- Add is orderable checkbox

### Closes :no_good_woman:
**[PT-1353](https://helsinkisolutionoffice.atlassian.net/browse/PT-1353):** 

## Screenshots :camera_flash:

Screenshot 2021-11-29 at 20 05 36

### DIFF
--- a/src/domain/event/eventForm/OrderableEventCheckboxField.tsx
+++ b/src/domain/event/eventForm/OrderableEventCheckboxField.tsx
@@ -4,24 +4,21 @@ import * as React from 'react';
 
 import CheckboxField from '../../../common/components/form/fields/CheckboxField';
 
-// This is its own component because how Formik recommends to handle situations
-// where form value is based on other form value.
-// see: https://github.com/formium/formik/issues/2204#issuecomment-574207100
-const VirtualEventCheckboxField = ({
+const OrderableEventCheckboxField = ({
   form,
   ...props
 }: FieldProps & CheckboxProps) => {
   const { setFieldValue } = form;
 
   const handleOnChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    setFieldValue('isVirtual', e.target.checked);
+    setFieldValue('isBookable', e.target.checked);
     if (e.target.checked) {
       setFieldValue('location', '');
-      setFieldValue('isBookable', false);
+      setFieldValue('isVirtual', false);
     }
   };
 
   return <CheckboxField {...props} form={form} onChange={handleOnChange} />;
 };
 
-export default VirtualEventCheckboxField;
+export default OrderableEventCheckboxField;
--- a/src/domain/occurrence/CreateOccurrencePage.tsx
+++ b/src/domain/occurrence/CreateOccurrencePage.tsx
@@ -38,7 +38,10 @@ import Container from '../app/layout/Container';
 import PageWrapper from '../app/layout/PageWrapper';
 import { ROUTES } from '../app/routes/constants';
 import ErrorPage from '../errorPage/ErrorPage';
-import { VIRTUAL_EVENT_LOCATION_ID } from '../event/constants';
+import {
+  BOOKABLE_TO_SCHOOL_LOCATION_ID,
+  VIRTUAL_EVENT_LOCATION_ID,
+} from '../event/constants';
 import {
   EDIT_EVENT_QUERY_PARAMS,
   NAVIGATED_FROM,
@@ -122,6 +125,9 @@ const CreateOccurrencePage: React.FC = () => {
 
         const venueData = data?.venue;
         const isVirtualEvent = event.location?.id === VIRTUAL_EVENT_LOCATION_ID;
+        const isBookable =
+          event.location?.id === BOOKABLE_TO_SCHOOL_LOCATION_ID;
+        const isLocationEditable = isVirtualEvent || isBookable;
         const eventName = omit(event.name, '__typename');
         const eventLangs = Object.entries(eventName).reduce<string[]>(
           (prev, [lang, value]) => (value ? [...prev, lang] : prev),
@@ -158,8 +164,9 @@ const CreateOccurrencePage: React.FC = () => {
             ? new Date(event.pEvent.enrolmentStart)
             : null,
           isVirtual: isVirtualEvent,
+          isBookable,
           neededOccurrences: event.pEvent.neededOccurrences || '1',
-          location: isVirtualEvent ? '' : event.location?.id || '',
+          location: isLocationEditable ? '' : event.location?.id || '',
           hasAreaForGroupWork: venueData?.hasAreaForGroupWork ?? false,
           hasClothingStorage: venueData?.hasClothingStorage ?? false,
           hasIndoorPlayingArea: venueData?.hasIndoorPlayingArea ?? false,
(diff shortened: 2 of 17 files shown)


src/domain/event/eventForm/OrderableEventCheckboxField.tsx
```diff
@@ -0,0 +1,27 @@
+import { FieldProps } from 'formik';
+import { CheckboxProps } from 'hds-react';
+import * as React from 'react';
+
+import CheckboxField from '../../../common/components/form/fields/CheckboxField';
+
+// This is its own component because how Formik recommends to handle situations
+// where form value is based on other form value.
+// see: https://github.com/formium/formik/issues/2204#issuecomment-574207100
```
Jokinen, line 9: Is this field based on other form values? I can tell it has side effects which affect other form values.

src/domain/event/eventForm/OrderableEventCheckboxField.tsx
```diff
@@ -0,0 +1,27 @@
...
+}: FieldProps & CheckboxProps) => {
+  const { value: isOrderable } = props.field;
+  const { setFieldValue } = form;
+
+  React.useEffect(() => {
+    if (isOrderable) {
+      setFieldValue('location', '');
+      setFieldValue('isVirtual', false);
+    }
+  }, [isOrderable, setFieldValue]);
```
Jokinen, line 22: Wondering about this--why not override the `onChange` handler? Pseudo:

```tsx
(e) => {
  const value = e.target.value;

  if (value) {
    setFieldValue('location', '');
    setFieldValue('isVirtual', false);
  }

  setFieldValue(e.target.name, value);
}
```

Seems like the usage of `useEffect` could result in situations where this component ends up syncing state when `isOrderable` is controlled in some other component. That seems like something that could make the logic of the form harder to understand when debugging.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -5,6 +5,7 @@ import * as React from 'react';
 import Modal from 'react-modal';
 import { toast } from 'react-toastify';
 
+import { createEmptyLocalizedObject } from '../../../constants';
```
Jokinen, line 8: Unused import.

src/domain/occurrence/utils.ts
```diff
@@ -26,17 +29,23 @@ export const getOccurrencePayload = ({
...
     startTime: values.startTime,
     endTime: values.endTime,
     languages: values.languages.map((lang) => ({ id: lang as Language })),
     pEventId,
-    placeId: isVirtual ? VIRTUAL_EVENT_LOCATION_ID : values.occurrenceLocation,
+    placeId: isVirtual
+      ? VIRTUAL_EVENT_LOCATION_ID
+      : isOrderable
+      ? ORDER_TO_SCROOL_LOCATION_ID
+      : values.occurrenceLocation,
```
Jokinen, line 48: I don't mind this pattern, but people often point out that it's hard to read.

Jokinen (commented):
Wasn't able to test yet because I haven't got permissions to access the admin, but I requested them. Otherwise looks good, minor code comments.

src/domain/event/eventForm/OrderableEventCheckboxField.tsx
Sartonon: Hmm I think it is not based on other form values but rather automatically changed to different value based on other values

Sartonon: But that is actually handled VirtualEventCheckboxField. Should be maybe moved into this component

src/domain/event/eventForm/OrderableEventCheckboxField.tsx
Sartonon: I'm afraid that this is done with useEffect because your example breaks Formiks internal validation logic.

Sartonon: I have had problems with validation going out of sync when trying to set multiple field values inside one event handler

Sartonon: I think I tried that also when I tried to solve this problem. But not necessarily in separate component like this one...

Jokinen: Good to know 👍

Jokinen (approved):
Tested the UI now and it seemed to be working 👍

No must fix comments on my end so this is good to me.

src/domain/occurrence/CreateOccurrencePage.tsx
```diff
@@ -38,7 +38,10 @@ import Container from '../app/layout/Container';
 import PageWrapper from '../app/layout/PageWrapper';
 import { ROUTES } from '../app/routes/constants';
 import ErrorPage from '../errorPage/ErrorPage';
-import { VIRTUAL_EVENT_LOCATION_ID } from '../event/constants';
+import {
+  ORDER_TO_SCROOL_LOCATION_ID,
```
nikomakela, line 42: typo

src/domain/event/eventForm/VirtualEventCheckboxField.tsx
```diff
@@ -11,16 +11,17 @@ const VirtualEventCheckboxField = ({
...
-  const { value: isVirtual } = props.field;
   const { setFieldValue } = form;
 
-  React.useEffect(() => {
-    if (isVirtual) {
+  const handleOnChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setFieldValue('isVirtual', e.target.checked);
+    if (e.target.checked) {
       setFieldValue('location', '');
+      setFieldValue('isOrderable', false);
```
nikomakela, line 20: Would bookable be a better term? I first thought this PR was something about reordering the checkboxes :)

src/domain/occurrence/locationFormPart/LocationFormPart.tsx
```diff
@@ -43,19 +44,28 @@ const LocationFormPart: React.FC<{ selectedLanguages: Language[] }> = ({
...
-            />
+            <div className={styles.locationCheckboxes}>
+              <Field
+                label={t('eventForm.location.labelEventIsHeldVirtual')}
+                name="isVirtual"
+                disabled={isOrderable}
+                component={VirtualEventCheckboxField}
+              />
+              <Field
+                label="Tapahtuma on tilattavissa päiväkotiin tai kouluun"
```
nikomakela, line 55: Needs translations

nikomakela (changes requested):
Would change a variable name, fix a typo and add translations

src/domain/event/eventForm/VirtualEventCheckboxField.tsx
Sartonon: I agree

src/domain/occurrence/locationFormPart/LocationFormPart.tsx
Sartonon: Ticket said that we aren't even sure if this is the correct label we are going to use so I didn't get translation for it yet

nikomakela (approved):
There are still some misleading "orderable" terms in tests and component name, but it's approvable

nikomakela (approved):
still looks good to me :)

PR #241, opened 2021-12-02
PT-1347 | Fix validation in edit enrolment page + add tests

Sartonon:
## Description :sparkles:

- Fix enrolment form validation to match with Teacher UI
- Take edited enrolment group size into account with remaining seat when validating group size

### Closes :no_good_woman:
**[PT-1347](https://helsinkisolutionoffice.atlassian.net/browse/PT-1347):** 

### DIFF
--- a/src/domain/enrolment/EditEnrolmentPage.tsx
+++ b/src/domain/enrolment/EditEnrolmentPage.tsx
@@ -9,13 +9,11 @@ import LoadingSpinner from '../../common/components/loadingSpinner/LoadingSpinne
 import {
   Language,
   NotificationType,
-  OccurrenceSeatType,
   useEnrolmentQuery,
   useUpdateEnrolmentMutation,
 } from '../../generated/graphql';
 import useHistory from '../../hooks/useHistory';
 import useLocale from '../../hooks/useLocale';
-import { assertUnreachable } from '../../utils/typescript.utils';
 import Container from '../app/layout/Container';
 import PageWrapper from '../app/layout/PageWrapper';
 import { ROUTES } from '../app/routes/constants';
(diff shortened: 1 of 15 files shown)


src/domain/enrolment/EditEnrolmentPage.tsx
```diff
@@ -145,10 +176,18 @@ const EditorEnrolmentPage: React.FC = () => {
...
+                <EnrolmentForm
+                  onSubmit={handleSubmit}
+                  initialValues={initialValues}
+                  minGroupSize={minGroupSize}
+                  maxGroupSize={maxGroupSize}
+                />
+              ) : (
+                <Notification label="Virhe" type="error">
+                  Virhe lomakkeen alustuksessa
+                </Notification>
```
Sartonon, line 189: I will ask for translatio if we are going to keep this here. Not 100% sure how we shoul to this kind of error handling, shouldn't if nothing strange happens in API

src/domain/enrolment/EditEnrolmentPage.tsx
```diff
@@ -157,4 +196,6 @@ const EditorEnrolmentPage: React.FC = () => {
   );
 };
 
+const isNumber = (n?: number | null): n is number => typeof n === 'number';
```
nikomakela, line 199: Lodash has the same, but acceptable.

src/domain/enrolment/EditEnrolmentPage.tsx
nikomakela: It's good to inform that something is not ok.

nikomakela (approved):
I tested the validation and didn't ran into any problems. There is also quite good amount of new tests related to this. Good job.

src/domain/enrolment/EditEnrolmentPage.tsx
Sartonon: Maybe I'll change  it

PR #245, opened 2021-12-16
PT-14137 | Delete beta texts

Sartonon:
### Closes :no_good_woman:

### DIFF
--- a/src/domain/app/i18n/sv.json
+++ b/src/domain/app/i18n/sv.json
@@ -4,7 +4,7 @@
       "permissionDeniedError": "Du har inte behörighet att utföra den här åtgärden"
     }
   },
-  "appName": "Kultus beta Admin",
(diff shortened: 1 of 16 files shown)


src/domain/app/i18n/sv.json
```diff
@@ -4,7 +4,7 @@
       "permissionDeniedError": "Du har inte behörighet att utföra den här åtgärden"
     }
   },
-  "appName": "Kultus beta Admin",
+  "appName": "Kultus",
```
nikomakela, line 7: Other languages still contains the removed "Admin" term

nikomakela (changes requested):
The "beta" term was there quite many times 😲 . "Admin" from Swedish is missing. After that, all seems to be well to be deployed!

PR #248, opened 2021-12-28
feat(profile): email can be given in my profile form

nikomakela:
PT-1411 PT-1417. Added person.emailAddress to my profile form (creation and update).

- In profile creation phase, the email field is auto populated from user.email.
- The user email is still shown as non-editable text
- A new person email address field is added and it's editable in both phases
- The new email input is required and validates the email format
- No changes to API was required after all

image

NOTE: the creation form is easy to check with unresponding or halted API ;)


src/domain/myProfile/CreateMyProfile.tsx
```diff
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next';
 import { toast } from 'react-toastify';
 
 import { useCreateMyProfileMutation } from '../../generated/graphql';
-import scrollToTop from '../../utils/scrollToTop';
+// import useHistory from '../../hooks/useHistory';
```
Sartonon, line 6: Can be deleted

src/domain/myProfile/myProfileForm/ValidationSchema.ts
```diff
@@ -5,6 +5,9 @@ import { VALIDATION_MESSAGE_KEYS } from '../../app/i18n/constants';
 export const getMyProfileValidationSchema = (type: 'create' | 'edit') => {
   const commonFormSchema = {
     name: Yup.string().required(VALIDATION_MESSAGE_KEYS.TIME_REQUIRED),
+    emailAddress: Yup.string()
+      .required(VALIDATION_MESSAGE_KEYS.TIME_REQUIRED)
```
Sartonon, line 9: Seems like `VALIDATION_MESSAGE_KEYS.TIME_REQUIRED` is used throughout this file. Maybe should be changed to avoid confusion :)

Sartonon (approved):
Looks good 👍

PR #249, opened 2021-12-31
Add autoAcceptanceMessage field

nikomakela:
PT-1400 PT-1435


src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -1232,6 +1232,103 @@ describe('enrolment type selector', () => {
...
+    act(() => userEvent.click(getFormElement('autoAcceptance')));
+    await waitFor(() => {
+      expect(getFormElement('autoAcceptanceMessage')).toBeInTheDocument();
+    });
+    userEvent.type(getFormElement('autoAcceptanceMessage'), 'bla bla');
+    userEvent.click(getFormElement('saveButton'));
+    await waitFor(() => {
+      expect(toastSuccess).toHaveBeenCalledWith('Tiedot tallennettu');
+    });
+    // TODO: Test that the formValues contains 'bla bla'
```
nikomakela, line 1284: - [ ] need to test that the submit process has the right values

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -1232,6 +1232,103 @@ describe('enrolment type selector', () => {
...
+          name: /mahdolliset lisätiedot vahvistusviestiin/i,
+        })
+      ).not.toBeInTheDocument();
+    });
+
+    userEvent.click(getFormElement('saveButton'));
+    await waitFor(() => {
+      expect(toastSuccess).toHaveBeenCalledWith('Tiedot tallennettu');
+    });
+    // TODO: Test that the formValues contains null for autoAcceptanceMessage
```
nikomakela, line 1328: - [ ] need to test that the submit process has the right values

nikomakela:
Also added the notifications requested in PT-1436
image

image

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -1232,6 +1233,148 @@ describe('enrolment type selector', () => {
   );
 });
 
+describe('auto acceptance for enrolments', () => {
+  it('renders auto acceptance message when auto accept is set on', async () => {
+    renderComponent({
+      mocks: [getEventMockedResponse({ autoAcceptance: false })],
+    });
+    // Wait for form to have been initialized
+    await screen.findByTestId('time-and-location-form');
```
Jokinen, line 1242: I wonder if we could detect the form's existence based on something human readable? For instance its title or a field we are going to interact with.

https://testing-library.com/docs/queries/bytestid/
> In the spirit of the guiding principles, it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names. Learn more about data-testids from the blog post "Making your UI tests resilient to change"

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -1232,6 +1233,148 @@ describe('enrolment type selector', () => {
...
+    // set as checked
+    act(() => userEvent.click(getFormElement('autoAcceptance')));
```
Jokinen, line 1244: I would expect that the `userEvent.click` helper is wrapped in an act and that the explicit `act` is not necessary here.

Possible changes that occur async should be flushed by `waitFor`. If this is not the case in practice, this approach is fine.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -1232,6 +1233,148 @@ describe('enrolment type selector', () => {
...
+      ],
+    });
+    // Wait for form to have been initialized
+    await screen.findByTestId('time-and-location-form');
+    // set as checked
+    act(() => userEvent.click(getFormElement('autoAcceptance')));
+    await waitFor(() => {
+      expect(getFormElement('autoAcceptanceMessage')).toBeInTheDocument();
+    });
+    userEvent.type(getFormElement('autoAcceptanceMessage'), 'bla bla');
```
Jokinen, line 1292: Wondering if the message you are using here is a kind of PoC message and whether you intended to align it with he usecase later? Up to you.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
```diff
@@ -1232,6 +1233,148 @@ describe('enrolment type selector', () => {
...
+        autoAcceptanceMessage: 'bla bla',
+      }),
+    });
+    expect(spyGetEditEventPayload).toHaveReturnedWith(
+      expect.objectContaining({
+        pEvent: expect.objectContaining({ translations: [] }),
+      })
+    );
+  });
+});
```
Jokinen, line 1376: It seems like you could combine the 3 test cases above into a single happy path testcase. I'd consider it as these are in essence integration level tests and this test file is very heavy as is.

src/domain/occurrence/utils.ts
```diff
@@ -144,20 +146,31 @@ export const getEditEventPayload = ({
       enrolmentEndDays: Number(enrolmentEndDays) || 0,
       neededOccurrences: Number(neededOccurrences) || 0,
       autoAcceptance,
+      translations:
+        autoAcceptance && autoAcceptanceMessage
+          ? [
+              {
+                languageCode: Language.Fi,
```
Jokinen, line 153: Seems odd that Finnish would be hardcoded here. Checking that it's intended.

Jokinen (commented):
General comments on the code.

Can't be tested yet.

src/domain/occurrence/utils.ts
nikomakela: it's intended until the UI has a multi language support. It's not designed yet and the ticket says that it's not needed yet. However, the API already supports it, because it would be harder to make the migrations afterwards. The API also uses FI as a fallback language.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
nikomakela: I don't mind - I just used this because it was used earlier also. I don't see anything wrong with this either. This form part starts with "tabs", but sure we could use first field of the first form section.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
nikomakela: I'm not sure if this part was here when you reviewed, but the thing is that I would like to test whether or not the value really changes in formik context and what is really posted to API.
```
    expect(spyGetEditEventPayload).toHaveBeenCalledWith({
      event: expect.anything(),
      formValues: expect.objectContaining({
        autoAcceptance: true,
        autoAcceptanceMessage: 'bla bla',
      }),
    });
    expect(spyGetEditEventPayload).toHaveReturnedWith(
      expect.objectContaining({
        pEvent: expect.objectContaining({
          translations: [
            { languageCode: 'FI', autoAcceptanceMessage: 'bla bla' },
          ],
        }),
      })
    );
```

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
nikomakela: No, I cannot, because I would like to test what is sent to API when the custom auto approvance message is not given and when it is given. Also the auto acceptance -checkbox matters here and it should clear the posted value when it's deactivated or otherwise I cannot clear the answer in database.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
nikomakela: I got warnings of it, but it may have been that I did not have the waitFor there next by then.

nikomakela: yep, the act was needless. Removed it.

src/domain/occurrence/__tests__/CreateOccurrencePage.test.tsx
nikomakela: the auto acceptance message is the whole beef here, so I'm not sure if I understood your comment properly :) Api changes are here (like mentioned in the start comment) https://github.com/City-of-Helsinki/palvelutarjotin/pull/243

nikomakela:
NOTE: API changes merged!

Jokinen (approved):
Tested the UI. It gets a ✅ from me.

There's a slight shift int he layout when the user toggles one the message compose textarea, but that's fine by me.

I'm also very slightly interested about the title and description of the notification. But as it's an admin facing UI, this approach is fine. For a customer facing application I would have checked how this title works for screen reader users.

For instance with this variation:

> Title: Vahvistusviestiin lisätään automaattisesti tietoja
> Description: Vahvistusviesti sisältää automaattisesti seuraavat tiedot: ...

The title can stand on its own in cases where screen readers use it to name the notification. Just a FYI for the future.

PR #257, opened 2022-01-24
fix: prevent fetching the places without place id

nikomakela:
PT-1470 PT-1471


Jokinen (commented):
`src/domain/occurrences/occurrencesTable/ActionsDropdown.tsx`

Seems to invoke this query as well. Is it problematic?

Tests haven't passed and Github is glitching when it comes to the log. Are those failures unrelated with these changes?

Sartonon (approved):
Looks good 👍

PR #267, opened 2022-04-13
chore: normalize price field value

nikomakela:
PT-1489.

Price formatter / normalizer: 
- Converts commas to dots (while writing) 
- Does not allow spaces (while writing) 

image


nikomakela:
The PR #268 is an alternative for this one. I'm waiting for feedback from the product owner, but I personally prefer the solution in this PR.

Jokinen:
Hmn, I'm thinking whether this is the UX we want to have 🤔

Whether the correct character is a comma or a period depends on locale. I think that the value the user sees should depend on their locale. E.g. in the screenshot we should be seeing a comma.

The value that gets sent to the server should be formatted in whichever way the server expects. I expected the value to get transformed before send, not on change.

nikomakela:
The product owner answered that we need to support a real free text like "make an offer" or "ask an offer", etc. So, closing this pull request and will use #268 instead.

Jokinen (approved):

PR #268, opened 2022-04-14
chore: convert price field to a free text field

nikomakela:
PT-1489 PT-1490. This branch is an alternative for #267 (where the price field number is formatted / normalized). This branch makes the field fully free text.


nikomakela:
Asked from product owner and we need to support a real free text like "ask an offer", so this PR should be used instead of #267.

I will add "ask an offer" as an example to the unit test.

- [ ]

Jokinen (approved):

melniiv (approved):

PR #269, opened 2022-04-20
PT-1496 | Add root navigation item as the first link

Jokinen:
## Description :sparkles:

Adds root page links as the first link in the dropdown so that the root page is more easily accessible.

### Closes :no_good_woman:

**[PT-1496](https://helsinkisolutionoffice.atlassian.net/browse/PT-1496)**

## Screenshots :camera_flash:

Screenshot 2022-04-20 at 13 57 21


nikomakela (approved):
Snapshot needs an update

PR #282, opened 2023-01-23
test: improve header tests fail safeness

nikomakela:


karisal-anders (approved):
Is the `act` import necessary? It doesn't look like it is used. If it isn't used IMHO it would be better to remove it as unnecessary. Otherwise LGTM.

PR #298, opened 2023-06-20
Update to Node 20 and upgrade and migrate all the dependencies

nikomakela:
PT-1674.

Upgraded all the dependencies with `yarn-update-all`. Installed Craco (SWC) to override some critical Create React App (CRA) configurations. Installed the SWC version, which makes the compiling and test running much faster than what it is with the Webpack.

The primary thing here is a Node update, because the old version was running out of lifecycle.
It's good to update the dependencies by time anyway, but also the Node update requires some of them to be updated.

The Helsinki Design System cannot be upgraded to the latest yet, because it does not support the CRA v5. Also, it should be noted that the CRA v5 and the whole CRA is currently deprecated and it is suggested that it should be replaced with the Vite.

When the Node and React were updated, also the Testing-library was a must to be updated. The newest version of the React-Testing-Library and Testing-library-User-Events needs the user events to be awaited. The new UserEvent library also contains a bug currently, which makes clearing or typing to any input hard. The act might not be waited so it might not get ever finished, which makes the unit tests to fail. Double clicking, pretyping, etc. helps there. Also it should be noted that jest.spy cannot override library methods if the library is imported from the index. In these cases, the whole lib must be mocked with jest.mock and jest.requireActual.

The new dependencies might have new or modified types, so the code must have been updated. The Apollo hooks works a bit differently and their cache features are hard to get mocked. Also the MSW mocks and Apollo mocks are more strict and they must have the mock for submits. The React Router was one that needed the changes to the code most. The history object is no longer used, but a navigate command from the useNavigate instead.

----

Node life cycle: https://endoflife.date/nodejs
Update all teh dependencies at once: https://yarnpkg.com/package/yarn-upgrade-all
Create React App: https://create-react-app.dev/. Note: already deprecated: https://github.com/reactjs/react.dev/pull/5487.
Styling with Sass -- Note: The “node-sass” -package is deprecated and it should be replaced with the Dart version of the Sass, a package called “sass”: https://www.npmjs.com/package/sass
For overriding the CRA configurations: https://craco.js.org/docs/
SWC-tools - Rust-based platform for the Web: https://swc.rs/
Use swc in your create-react-app with craco for faster compilation, development and tests: https://www.npmjs.com/package/craco-swc
----
Upgrading the React-Router

Migration guide: https://reactrouter.com/en/main/upgrading/v5#backwards-compatibility-package
Biggest changes, that needs lots of changes on our end:
History object is no longer used. Use navigate instead. Lots of mocking in Jest unit tests.
Routing does not support RegExp-paths. Routing is what controls our navigation inside the app – What URL renders what component/page.
----
Helsinki Design System

https://hds.hel.fi/getting-started/developer
We always must support the latest version of the HDS.
NOTE: The latest version of HDS does not fit to CRA based application. After 2.10.0 the styles are loaded differently.
----
React Testing Library and User-events

About React Testing Library: https://testing-library.com/docs/react-testing-library/intro/
About User-Events upgrading: 
https://www.npmjs.com/package/@testing-library/user-event. 
https://testing-library.com/docs/ecosystem-user-event/. 
Only the latest version is supported, but the user events has some bugs:
Clear not working: https://github.com/testing-library/user-event/discussions/970
Type not working: https://github.com/testing-library/user-event/issues/930


nikomakela:
There is something weird going on in the 3rd page (the event publishing page) of the event creation, because the mouse scrolling seems to stop working. Scrolling works with keyboard.

- [ ] Fix event creation 3rd page scrolling issue

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr298.dev.hel.ninja** :rocket::rocket::rocket:

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # ===============================================
-FROM helsinkitest/node:14-slim as appbase
+FROM helsinkitest/node:18-slim as appbase
```
melniiv, line 2: 18 or 20?

melniiv (approved):

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr298.dev.hel.ninja** :rocket::rocket::rocket:

Dockerfile
nikomakela: It's 18 until we have helsinkitest/node:20 available. 18 should work ok.

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr298.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr298.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr298.dev.hel.ninja** :rocket::rocket::rocket:

nikomakela:
The review server responds with an internal server error. This should not be merged before it's some how fixed.

- [ ] Fix the review server

PR #314, opened 2023-12-11
Fix enrolment edit link and form data initialization

nikomakela:
PT-1706

### The errors

The errenous section of the site:
image

1. When the "Muokkaa"  /edit button was clicked, the locale was added twice and the page was rendered as an empty grey page

image

Solution: Remove the explicit locale set from the link path.

2. When the "Muokkaa" /edit was clicked from the dropdown menu, the locale was not added twice, but another error which was related to the form data initialization was thrown

image

Solution: This was a special case and the error was always written in Finnish, when there were no audience limits set for a study group.

### The page that should have been shown and is now shown after the fixes

image

----

EDIT: There was also a 3rd error which was in the group size validation. There were some default limits which was a bad practise. Now, the min group size is set to 1 and maximum can be left undefined or it's calculated from the remaining space left. IT should als obe noted that this is an admin form -- not for public users.

### DIFF
--- a/src/domain/enrolment/EditEnrolmentPage.tsx
+++ b/src/domain/enrolment/EditEnrolmentPage.tsx
@@ -170,18 +170,12 @@ const EditorEnrolmentPage: React.FC = () => {
                 {t('enrolment.editEnrolmentBackButton')}
               </BackButton>
               <h1>{t('enrolment.editEnrolmentTitle')}</h1>
-              {!!minGroupSize && !!maxGroupSize ? (
-                <EnrolmentForm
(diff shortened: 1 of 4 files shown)


terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr314.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr314.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):

PR #328, opened 2024-02-23
PT-1725 | fix: fix production styles & running docker compose

karisal-anders:
### fix: fix production styles & running docker compose

Docker compose related:
 - Fix running docker compose by copy-pasting .env.development into .env
   because docker compose build args doesn't support any other files
   except .env, env_file setting affects environment variables in docker
   compose but it does not affect build args.
 - So in order to support docker compose build args and environment
   variables being the same they have to either be put them into .env
   (no other file will suffice, only .env) or e.g. set them using a
   script to the environment before running docker compose.

Production styles related:
 - Fix production styles by marking all styles in header.module.scss
   as !important and --footer-background in footer.module.scss as
   !important.
   - This is a workaround to make sure styles are applied even if CSS
	 order varies. Switching from Craco to Vite might remove the need
	 for this hack. Tried disabling CSS chunking in webpack but either
	 it didn't work or if it did work it didn't help.

refs PT-1725

### Related :handshake:

[PT-1725](https://helsinkisolutionoffice.atlassian.net/browse/PT-1725)

[PT-1725]: https://helsinkisolutionoffice.atlassian.net/browse/PT-1725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,7 +1,7 @@
 @import './vendor/normalize';
-
 @import 'fonts';
 @import 'layout';
+@import 'variables';
 
 html {
   height: 100%;
--- a/.env
+++ b/.env
@@ -14,3 +14,19 @@ REACT_APP_SENTRY_DSN=
 REACT_APP_ENVIRONMENT=local
 REACT_APP_APPLICATION_NAME=$npm_package_name
 REACT_APP_VERSION=$npm_package_version
+
+# Environment variables for the development environment:
+#
+# Copied here from .env.development because docker compose build args doesn't support
+# any other files except .env, env_file setting affects environment variables in docker
+# compose but it does not affect build args.
+#
+# So in order to support docker compose build args and environment variables being the
+# same they have to either be put them into .env (no other file will suffice, only .env)
+# or e.g. set them using a script to the environment before running docker compose.
+REACT_APP_API_URI=https://kultus.api.test.hel.ninja/graphql
+REACT_APP_API_REPORT_URI=https://kultus.api.test.hel.ninja/reports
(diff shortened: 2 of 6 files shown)


terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr328.dev.hel.ninja** :rocket::rocket::rocket:

src/styles/main.scss
```diff
@@ -1,7 +1,8 @@
+@import '~hds-design-tokens/lib/all.scss';
```
nikomakela, line 1: I think this won't be needed. I think it duplicates the tokens. I would remove the whole `hds-design-tokens` dependency, because if it is not used, I don't think it is even included in the final bundle.

nikomakela (approved):
Please check the hds-design-tokens usage. Otherwise OK.

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr328.dev.hel.ninja** :rocket::rocket::rocket:

src/styles/main.scss
karisal-anders: The breakpoints are still used through hds-design-tokens, including them from hds-core's :root we'd get some extra styles that we don't necessarily want to include. Not removing hds-design-tokens in this PR, if a suitable solution is found later and there's time and inclination this can be improved.

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr328.dev.hel.ninja** :rocket::rocket::rocket:

PR #363, opened 2024-10-30
PT-1807 | Upgrade @apollo/client, HDS, RHHC, Vite related packages & remove warnings from tests

karisal-anders:
## Description :sparkles:

Upgrade packages:
 - @apollo/client to ^3.10.1, and HDS & RHHC to latest
 - all vite related packages to latest

Remove warnings from tests

### Related :handshake:

- [PT-1807](https://helsinkisolutionoffice.atlassian.net/browse/PT-1807)
- Continuation in PR #367 

## Testing :alembic:

Please use node v20 to test (See [Dockerfile](https://github.com/City-of-Helsinki/palvelutarjotin-admin/blob/PT-1807-upgrade-and-cleanup/Dockerfile#L2)).

### Automated tests :gear:️

Tests [ran](https://dev.azure.com/City-of-Helsinki/kultus/_build/results?buildId=208881&view=logs&jobId=8267e077-3a31-5bba-93c5-4b3e55896ad0&j=8267e077-3a31-5bba-93c5-4b3e55896ad0&t=f8bdbb17-be9a-593e-dc0b-831a1df4a0a7) in CI/CD pipeline without warnings except:
```
One of your dependencies, babel-preset-react-app, is importing the
"@babel/plugin-proposal-private-property-in-object" package without
declaring it in its dependencies. This is currently working because
"@babel/plugin-proposal-private-property-in-object" is already in your
node_modules folder for unrelated reasons, but it may break at any time.

babel-preset-react-app is part of the create-react-app project, which
is not maintianed anymore. It is thus unlikely that this bug will
ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to
your devDependencies to work around this error. This will make this message
go away.
```
which **won't be handled** in this PR, but as part of PR #367 

## Additional notes :spiral_notepad:

[PT-1807]: https://helsinkisolutionoffice.atlassian.net/browse/PT-1807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #366, opened 2024-11-08
fix: idle timer provider

nikomakela:
Fixes the unit test issue in #363  .

PT-1807.

The onIdle was executed too fast when running the unit tests. The logout from the useOidcClient was not yet initialized, but the onIdle action was already called, so the function failed to execute.


karisal-anders:
> The problem was in environment variable. The idle timeout had the comments included `VITE_APP_IDLE_TIMEOUT_IN_MS=3600000 # 60 * 60 * 1000ms = 60min`.

Ok, good to know, I can take that into account and fix the environment variable setting in #363.

PR #378, opened 2025-03-19
Persisted apollo cache

nikomakela:
PT-1846.

Create and initialize the Apolo clients (Kukkuu API and Headless CMS) in a more consistent way. Handle the cache in a way that if a new client is created, the old existing cache is used. Also bring a new useApolloClient hook for both the clients.

Added CMS URL rewriting transformation link to the CMS Apollo client to convert internal CMS links to internal Kultus app links.

The utility `queryPageWithUri` is now moved to a new `services.ts` and needs client as param. A function that changes a state or fetches data or more like a service function instead of utility. Since the CMS Apollo client is refactored, it's better to give the client as a parameter instead of trying to import it or create a new one, because then we would not have any cache initialized for the client.

Copied the Apollo cache persistor implementation from the Kukkuu UI
(https://github.com/City-of-Helsinki/kukkuu-ui/pull/662):
- Installed apollo3-cache-persist.
- Implemented TimedApolloCachePersistor that creates a new persisted Apollo cache persistor with an expiration time. This class automatically clears expired caches and provides methods for persisting and restoring the cache to/from localStorage.

Prepopulated the cache for languages query from hardcoded responses and use  theReact-Router Link with HCRC-lib.


src/headless-cms/apollo/apolloClient.ts
```diff
@@ -0,0 +1,230 @@
...
+    if (apolloCachePersistor.hasPersistedCacheExpired()) {
+      // eslint-disable-next-line no-console
+      console.info('Persisted cache has expired.');
+      await apolloCachePersistor.purge();
+    } else {
+      await apolloCachePersistor.restore();
+      // eslint-disable-next-line no-console
+      console.info('Persisted cache has been restored.');
+    }
+  })();
```
nikomakela, line 108: Note this racing condition. Sometimes the cache is restored fast enough, but sometimes not. However, the only issue should be that a needless CMS query is been made. I cannot change the Apolloclient to be a promise, because we cannot await for it everywhere were it's needed. It should be a static variable that is always available.

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr378.dev.hel.ninja** :rocket::rocket::rocket:

src/domain/app/apollo/__tests__/apolloClient.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    expect(mergedState.ROOT_QUERY!.existingField).toEqual('existingValue');
+    // In array, the source array is initial state and then the new values are appended.
+    expect(mergedState.ROOT_QUERY!.arrayField).toEqual([
+      { __typename: 'ArrayFieldType', value: 1 },
+      { __typename: 'ArrayFieldType', value: 2 },
+      { __typename: 'ArrayFieldType', value: 4 },
+      { __typename: 'ArrayFieldType', value: 3 },
+      { __typename: 'ArrayFieldType', value: 5 },
+    ]);
+  });
```
karisal-anders, line 104: How about simplifying a bit and directly asserting the equality to the whole cache output?
```suggestion
    const client = initializeCMSApolloClient(existingState);
    initializeCMSApolloClient(initialState);

    const mergedState = client.cache.extract();
    expect(mergedState).toEqual({
      ROOT_QUERY: {
        __typename: 'Query',
        existingField: 'existingValue',
        initialField: 'initialValue',
        existingEmbeddedField: {
          __typename: 'EmbeddedType',
          embeddedField: 'embeddedValue',
          staticEmbeddedField: 'staticEmbeddedValue',
          newEmbeddedField: 'newValue',
        },
        arrayField: [
          { __typename: 'ArrayFieldType', value: 1 },
          { __typename: 'ArrayFieldType', value: 2 },
          { __typename: 'ArrayFieldType', value: 4 },
          { __typename: 'ArrayFieldType', value: 3 },
          { __typename: 'ArrayFieldType', value: 5 },
        ],
      },
    });
```

I intentionally broke the data and tested that the equality testing worked—it gave a human readable output of the non-matching field:
```
 FAIL  src/headless-cms/apollo/__tests__/apolloClient.test.ts > initializeCMSApolloClient > should restore the cache with merged existing and initial states
AssertionError: expected { ROOT_QUERY: { …(5) } } to deeply equal { ROOT_QUERY: { …(5) } }

- Expected
+ Received

  Object {
    "ROOT_QUERY": Object {
      "__typename": "Query",
      "arrayField": Array [
        Object {
          "__typename": "ArrayFieldType",
          "value": 1,
        },
        Object {
          "__typename": "ArrayFieldType",
          "value": 2,
        },
        Object {
          "__typename": "ArrayFieldType",
          "value": 4,
        },
        Object {
          "__typename": "ArrayFieldType",
          "value": 3,
        },
        Object {
          "__typename": "ArrayFieldType",
          "value": 5,
        },
      ],
      "existingEmbeddedField": Object {
-       "__typename": "EmbeddedTyped",
+       "__typename": "EmbeddedType",
        "embeddedField": "embeddedValue",
        "newEmbeddedField": "newValue",
        "staticEmbeddedField": "staticEmbeddedValue",
      },
      "existingField": "existingValue",
      "initialField": "initialValue",
    },
  }

 ❯ src/headless-cms/apollo/__tests__/apolloClient.test.ts:80:25
     78|
     79|     const mergedState = client.cache.extract();
     80|     expect(mergedState).toEqual({
       |                         ^
     81|       ROOT_QUERY: {
     82|         __typename: 'Query',
```

src/domain/app/apollo/apolloClient.ts
```diff
@@ -1,140 +1,193 @@
...
+  const storeRef = useRef<ApolloClient<NormalizedCacheObject>>();
+  if (!storeRef.current) {
+    storeRef.current = initializeApolloClient(initialApolloState);
+  }
+  return storeRef.current;
+}
+
+/**
+ * Reset the global variable for Apollo Client.
+ * NOTE: Helps in unit tests when running multipel tests simultaneously.
```
karisal-anders, line 163:
```suggestion
 * NOTE: Helps in unit tests when running multiple tests simultaneously.
```

src/domain/app/apollo/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+} from '@apollo/client';
+const excludeArgs =
+  (excludedArgs: string[]) =>
+  (
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    args: Record<string, any> | null
+  ) =>
+    args
+      ? Object.keys(args).filter((key: string) => !excludedArgs.includes(key))
+      : false;
```
karisal-anders, line 14: Thought I'd refactor this moved function to be faster with the use of Set (it was O(n*m) before) and now it should be much faster on average, and also make the code more readable at a glance by using the keyword `function` here to make the nesting of functions more readily apparent than it is when using nested arrow functions:
```suggestion

function excludeArgs(excludedArgs: string[]) {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  return function (args: Record<string, any> | null) {
    const excludedArgsSet = new Set(excludedArgs);
    return args
      ? Object.keys(args).filter((key: string) => !excludedArgsSet.has(key))
      : false;
  };
}
```

karisal-anders: Looking at the code this is not used much, and is only used with ['page'] parameter so not much gain here. But, a bit of readability, and the code change does not add much complexity to the code either, so the tradeoff is okay IMHO.

src/domain/app/footer/__tests__/__snapshots__/Footer.test.tsx.snap
```diff
@@ -76,7 +76,7 @@ exports[`Footer matches snapshot 1`] = `
             © 
             Copyright
              
-            2024
+            2025
```
karisal-anders, line 79: The test would be good to fix in a long term way so that this test wouldn't break come 2026. Not necessarily in this PR though.

src/domain/app/layout/__tests__/__snapshots__/PageLayout.test.tsx.snap
```diff
@@ -651,7 +651,7 @@ exports[`PageLayout matches snapshot 1`] = `
               © 
               Copyright
                
-              2024
+              2025
```
karisal-anders, line 654: The test would be good to fix in a long term way so that this test wouldn't break come 2026. Not necessarily in this PR though.

src/headless-cms/apollo/__tests__/apolloClient.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    initializeCMSApolloClient(initialState);
+
+    expect(extractSpy).toHaveBeenCalled();
+    expect(restoreSpy).toHaveBeenCalledWith(
+      expect.objectContaining(initialState)
+    );
+  });
+
+  it('should restore the cache with merged existing and initial states', () => {
```
karisal-anders, line 42: Same comment here as in https://github.com/City-of-Helsinki/palvelutarjotin-admin/pull/378/files#r2005059243

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr378.dev.hel.ninja** :rocket::rocket::rocket:

PR #394, opened 2025-04-11
feat: replace Redux context with React context implementation

nikomakela:
PT-1904.

Get rid of Redux and replace it with React context implementation. The active organisation was handled with a Redux store that was persisted in the local storage. Now the same is done with React Context.


terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr394.dev.hel.ninja** :rocket::rocket::rocket:

src/domain/organisation/contextProviders/OrganisationStoragePersistor.tsx
```diff
@@ -0,0 +1,58 @@
...
+   * @example
+   * OrganisationStoragePersistor.save('12345');
+   */
+  static save(organisationId: string): void {
+    // eslint-disable-next-line no-console
+    console.debug('OrganisationStoragePersistor:save', organisationId);
+    localStorage.setItem(
+      organisationLocalStorageKey,
+      JSON.stringify(organisationId)
+    );
```
karisal-anders, line 20: Wouldn't it be good to prefix the storage key with the profile ID / user ID here, as now the same key is shared for multiple users?

src/domain/organisation/contextProviders/useStoredOrganisationState.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Fetches the user's profile data, specifically their organisations, when the user is authenticated.
+ *
+ * @returns The result of the `useMyProfileQuery` hook.
+ */
+function useProfileOrganisationsQuery() {
+  const { isAuthenticated } = useOidcClient();
+  const isLoggedIn = !isAuthenticated(); // Skip when NOT logged in
+  return useMyProfileQuery({ skip: isLoggedIn });
```
karisal-anders, line 17: It looked like the comment and code variable name were out-of-sync. How about this?
```suggestion
  const isNotLoggedIn = !isAuthenticated();
  return useMyProfileQuery({ skip: isNotLoggedIn });
```

src/domain/organisation/contextProviders/OrganisationStoragePersistor.tsx
nikomakela: I don't think it gives any value. The OIDC key itself is not persisted like that either. Also, the old implentation was not done like that. I think it would make things more complicated, but does not bring anything more. When a new login is made, the organisation selection is "random", since it just uses the first organistaion in a list of organisations, which are in random order. If the organisation is picked from local storage, it is still validated against the organisations that are set for profile. Also, the last validation should be done by the API, whether or not the client has permissions to manage the organisation.

nikomakela: Actually, I even started thinking that I could remove the environment and app specific key too, since the organisation id will be unique in API anyway, so it cannot match in different apps and envs. We should not collect a list of organisations in local storage of a browser.

karisal-anders: Why remove the app specific key? For sure the organisation ID is related to this application and not some other, and just to make it clear that it doesn't possibly overlap with some other app's data it'd be justified to use at least an app specific key here.

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr394.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr394.dev.hel.ninja** :rocket::rocket::rocket:

PR #401, opened 2025-04-25
docs: improve readme

nikomakela:
PT-1880.

Add more content to README. Use [Kukkuu UI](https://github.com/City-of-Helsinki/kukkuu-ui) as a base for content and style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ Providers' UI - A restricted administration client for Kultus event providers (f
     - [Environment configuration](#environment-configuration)
       - [Running using local Node.js](#running-using-local-nodejs)
       - [Running using Docker](#running-using-docker)
-      - [Running the Kultus backend locally](#running-the-kultus-backend-locally)
-      - [Running Tunnistamo](#running-tunnistamo)
+      - [Running Tunnistamo as a local authorization service](#running-tunnistamo-as-a-local-authorization-service)
   - [Husky Git Hooks](#husky-git-hooks)
     - [Pre-commit Hook](#pre-commit-hook)
     - [Commit-msg Hook](#commit-msg-hook)
@@ -194,7 +193,7 @@ This project is built using the following key frameworks and libraries:
 ### Getting started
 
 1. Clone this repository
-2. If you're new to multiple `.env*` files, read Next.js's [Environment Variable Load Order](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#environment-variable-load-order)
+2. If you're new to multiple `.env*` files, read Vite's [Env variables](hhttps://vite.dev/guide/env-and-mode#env-files)
 3. If you're new to multiple Docker compose files, read Docker's [Merge Compose Files](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/)
 4. Follow the instructions below for your preferred way of running this app:
    - [Running using local Node.js](#running-using-local-nodejs) **or**
@@ -225,7 +224,7 @@ There are multiple types of events:
 
 - **Multiple occurrences might be needed**. When an event is created (in Kultus Provider's UI), the admin can choose whether an enrolment needs multiple occurrences to be chosen. In such events, the registrant must commit to attending the event on multiple event dates. These enrolments are always handled in Kultus internally and are automatically approved.
 
-- **Enrolments queue** might also be enabled, which means that when an occurrence is fully booked, the study groups can enrol to queue. IF any enrolment cancellations are made or the enrolments are reorganizing by the admins, enrolments in queue can be promoted to the group of actual registrations.
+- **Enrolments queue** might also be enabled, which means that when an occurrence is fully booked, the study groups can enrol to queue. IF any enrolment cancellations are made or the enrolments are reorganized by the admins, enrolments in queue can be promoted to the group of actual registrations.
 
 #### Environment configuration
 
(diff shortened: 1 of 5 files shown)


terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr401.dev.hel.ninja** :rocket::rocket::rocket:

README.md
```diff
@@ -18,46 +30,242 @@ Staff interface for Palvelutarjotin
...
+- help enrolment process
+- help providers (culture and leisure department and partly metropolitan-area ecosystem) to show activities
+- gather data (equal possibilities and use to all group, use of Kuva´s work, resource needs)
+
+**Roles**
+
+- Educator can be school or kindergarden teacher who seeks educational activities for study groups.
+- Provider is third party user or user from KUVA services who wants to provide activities for educators
+- Admin manages rights and pages, modifies front page, categories and newsletter, statistics
+
```
karisal-anders, line 62: It would be IMHO better to somehow make this section specifically talk about what this repository is about, not what Kultus is about in general as a whole. Or at least add the specific part to this and have a the general level stuff in addition to the specific part. Now this section does not really talk about what this repository is about.

karisal-anders (approved):
It would be beneficial IMHO to make the "About" section more specifically about this repository's service, not just Kultus in general, or in addition to the general level information. The graph could probably be improved (i.e. it is very full of links and they become hard to read because there are so many of them), and the wordings / terms could be consolidated (i.e. using one term to mean one thing, not two terms to mean one thing, e.g. educator/teacher → teacher) for clarity etc.

BTW looking at the graph in Github it looked more readable than when looking at it using Pycharm's Jetbrains official Mermaid plugin. So YMMV depending on the used Mermaid graph renderer.

So, in a nutshell: the about section would be good to update now, but otherwise let's go with this now, can be iterated more later.

Graph in Github:
- graph-github

Graph in Pycharm:
- graph-pycharm

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr401.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr401.dev.hel.ninja** :rocket::rocket::rocket:

docs/setup-tunnistamo.md
```diff
@@ -0,0 +1,51 @@
...
+- Authorization callback URL: http://tunnistamo-backend:8000/accounts/github/login/callback/
+
+Save. You'll need the created **Client ID** and **Client Secret** for configuring tunnistamo in the next step.
+
+## Install local tunnistamo
+
+Clone https://github.com/City-of-Helsinki/tunnistamo/.
+
+Follow the instructions for setting up tunnistamo locally. Before running `docker-compose up` set the following settings
+in tunnistamo roots `docker-compose.env.yaml`:
```
karisal-anders, line 24: There's no `docker-compose.env.yaml` so this part is out of date.

README.md
```diff
@@ -1,103 +1,441 @@
...
+  3. Events that have external enrolments. Then Kultus does not know the enrolment status and cannot do any management to them.
+
+- **Enrolments can be automatically or manually approved.** IF the enrolments are approved automatically, the approval is given immediately after the enrolment is sent and the event still had enough space left for the study group that enrolled. If the enrolments are approved manually, the enrolments are managed by the event admins or providers through the Provider's UI.
+
+- **Multiple occurrences might be needed**. When an event is created (in Kultus Provider's UI), the admin can choose whether an enrolment needs multiple occurrences to be chosen. In such events, the registrant must commit to attending the event on multiple event dates. These enrolments are always handled in Kultus internally and are automatically approved.
+
+- **Enrolments queue** might also be enabled, which means that when an occurrence is fully booked, the study groups can enrol to queue. IF any enrolment cancellations are made or the enrolments are reorganizing by the admins, enrolments in queue can be promoted to the group of actual registrations.
```
karisal-anders, line 227: Small grammar fix:
```suggestion
- **Enrolments queue** might also be enabled, which means that when an occurrence is fully booked, the study groups can enrol to queue. IF any enrolment cancellations are made or the enrolments are reorganized by the admins, enrolments in queue can be promoted to the group of actual registrations.
```

README.md
```diff
@@ -1,103 +1,441 @@
...
+3. To build and start a docker container, you can use
+    **a development build:**
+    > NOTE: [compose.yml](./compose.yml) maps some local directories as a volume in order to get a hot reload for the code to work.
+    ```shell
+    DOCKER_TARGET=development docker compose up --build
+    ```
+
+    or 
+
+    if you want to run **a production build**, you most likely first need to unmap the composer volumes from [compose.yml](./compose.yml), by removing or fully commenting out the following section:
```
karisal-anders, line 347: Another way to make the development and production builds both work with Docker compose locally would be to use compose.yml and compose.override.yml files to achieve that (Haven't thought out the exact details in my head, but I think it'd probably be doable that way). The upside would be that user wouldn't have to manually change compose.yml which is committed to the repository and will therefore show up as changes. Not something necessary to do in this PR though.

README.md
karisal-anders: Looking now at the README as it starts with `Providers' UI` which ties to the `Provider` term in `Who use it` section of `About Kultus` there's an apparent connection to what this repository is about. Because of this I don't think the "About Kultus" section needs to be changed anymore, this connection makes this repository's function clear.

karisal-anders (approved):
Please take a look at the comments I left.

README.md
nikomakela: I think that using override-file is not a good practise in this kind of cases. I think that it should be left for those rare cases when you really need to override something in your local setup In cases when you need important local changes, but don't want to change any files in repository. I would prefer using [profiles](https://docs.docker.com/compose/how-tos/profiles/) or another named configuration file (e.g compose-production-build.yaml) instead. But, we do already have a DOCKER_TARGET variable for almost the same purpose.

nikomakela: I'm also thinking, that we should not provide a docker build stage for development purposes at all. We _don't have to_ support that kind of special cases, eventhough we can do that. We don't need to provide a working docker environment for development use and we can never provide all the tools anyway. Tools like editors and such with all their settings will always be something personal. It's easier, faster, simpler and convenient way to just install a node in your local computer, when you develop a node application. Whether somebody wants to do that in a docker container, he can do that, but we don't need to provide support for that.

nikomakela: So, I would be happy to remove the develop stage from dockerfile, remove the volumes mapping totally and make the production build stage the default stage.

terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr401.dev.hel.ninja** :rocket::rocket::rocket:

README.md
karisal-anders: > So, I would be happy to remove the develop stage from dockerfile, remove the volumes mapping totally and make the production build stage the default stage.

Please, let's discuss this live.

PR #403, opened 2025-05-06
fix: always navigate to root after organisation select

nikomakela:
PT-1925.


terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr403.dev.hel.ninja** :rocket::rocket::rocket:

PR #409, opened 2025-05-19
fix: remove process.env from vite config define

nikomakela:
PT-1927.

The `process.env` is not used anywhere, (except in Playwright).

Fixes the warning in startup saying:

```
$ vite preview
The `define` option contains an object with "PATH" for "process.env"
key.
It looks like you may have passed the entire `process.env` object to
`define`,
which can unintentionally expose all environment variables.
This poses a security risk and is discouraged.
```

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -11,9 +11,6 @@ const viteConfig = ({ mode }) => {
   process.env = { ...process.env, ...loadEnv(mode, process.cwd()) };
 
   return defineConfig({
-    define: {
-      'process.env': process.env,
-    },
     resolve: {
       alias: {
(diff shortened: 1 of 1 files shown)


terovirtanen:
**kultus-admin-ui branch is deployed to platta: https://kultus-admin-ui-pr409.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):
> The process.env is not used anywhere, (except in Playwright).

Environment variable handling will need some work with Playwright still—it's ok to remove this. Just saying that I've run into problems with import.meta.env & Playwright tests in development already, and haven't yet come up with a working solution to that.

